PR #140, opened 2026-07-28
fix(editor): reset a split tail to its own track, not the whole one (#131)

### LINKED ISSUE #131: bug: "Reset to Strava" on a split tail restores the whole pre-split track
### What

A split tail's "original" geometry snapshot is the **entire pre-split track**, not the tail's own portion. Pressing *Reset to Strava* on a tail therefore expands it back into a duplicate of the whole track it was cut out of, overlapping its sibling head on the map.

The label is also wrong on its own terms: a tail is a LOCAL row (synthetic negative id, `manual=True`) that never came from Strava, so there is no Strava original to reset to.

### Why

`split_activity` seeds the new tail row with the head's full pre-split geometry so the tail's time apportioning has the right denominator — `src/project/repo_activities.py:340-341`:

```python
tail.summary_polyline = head.summary_polyline
tail.elevation_profile_json = head.elevation_profile_json
```

`_write_track_geometry` then snapshots whatever is on the row on the first edit, and a fresh tail always has `is_edited == False` — `src/project/repo_activities.py:112-114`:

```python
if not row.is_edited:
    row.original_polyline = row.summary_polyline
    row.original_elevation_profile_json = row.elevation_profile_json
```

So `tail.original_polyline` ends up as the full pre-split track. `reset_activity_track` (`src/project/repo_activities.py:166`) restores it verbatim, and additionally scales `moving_time` / `elapsed_time` back **up** by the restored-to-edited distance ratio, so the tail also inherits the full track's duration.

Nothing gates the action: the editor shows *Reset to Strava* for anything with `is_edited == true` (`flutter_client/lib/src/projects/activity_editor_page.dart:90`), which every tail is, and neither `POST /activities/{id}/reset` nor `reset_activity_track` rejects a negative (local) id.

### Repro

1. Open an activity's track editor, *Split here* (or *Cut & add transport*) at some point.
2. Open the resulting tail piece — `<name> (2/2)` — in the track editor.
3. Press *Reset to Strava*.
4. The tail now carries the whole pre-split track, duplicating the head's geometry, with the full track's distance and times.

### Possible directions

- Hide *Reset to Strava* for local activities (`id < 0`) and reject the endpoint for them — a tail has no upstream original, so the action is meaningless.
- Or snapshot the tail's **own** geometry at creation, so reset undoes only edits made to the tail after the split, and keep the button. This keeps the seeding trick for time apportioning but decouples it from the snapshot.
- Either way, consider relabelling to something not Strava-specific for local pieces.

### Notes

Found while planning #127 (track edits not compounding with Split / Cut). Pre-existing on `main` and not caused by that fix; #127 only changes *which* full track gets snapshotted (edited rather than stored), so the behaviour here is unchanged by it.

rui-nar:
Fixes #131.

## The bug

`split_activity` seeds the new tail row with the head's **full pre-split geometry** so the tail's time apportioning has the right denominator (`repo_activities.py:356-357`). `_write_track_geometry` then snapshots whatever sits on the row into `original_*` on the first edit, and a fresh tail always has `is_edited == False` — so the tail's "original" ended up being the *entire* pre-split track.

Pressing **Reset to Strava** on a tail therefore expanded it back into a duplicate of the track it was cut out of, overlapping its sibling head on the map, and scaled its `moving_time` / `elapsed_time` up to the full track's duration.

## The fix

Direction 2 from the issue — snapshot the tail's **own** geometry at creation, rather than hiding the action:

```python
self._write_track_geometry(tail, tail_points)
tail.original_polyline = tail.summary_polyline
tail.original_elevation_profile_json = tail.elevation_profile_json
```

The seeding trick is kept for apportioning but decoupled from the snapshot, so Reset stays useful on a tail: it undoes edits made **after** the split, back to the piece the split produced. The head is untouched — it *is* the Strava activity, so it still snapshots the full pre-split track.

The label was wrong on its own terms too: a tail is a LOCAL row (synthetic negative id, `manual=True`) that never came from Strava. Local activities now get **"Reset track"** instead of **"Reset to Strava"**.

## Tests

Three API regressions in `tests/test_activity_split_api.py`, all of which **fail on the pre-fix code** (verified by reverting `repo_activities.py` and re-running):

- `test_split_tail_snapshots_its_own_geometry` — the tail's snapshot is its own track; the head's is still the full pre-split one.
- `test_reset_on_a_fresh_tail_is_a_noop` — geometry, distance and times unchanged, and the two pieces still tile the original track rather than overlapping.
- `test_reset_on_an_edited_tail_restores_the_post_split_track` — edit the tail, reset, get the post-split track back.

Plus `a local activity gets the non-Strava reset label` in `flutter_client/test/activity_editor_page_test.dart`.

All green: **1118 Python**, **603 Flutter**, `flutter analyze` clean.

## Out of scope

Resetting the **head** after a split restores the full Strava track while the tail still exists, which duplicates geometry the same way. That is arguably what "Reset to Strava" promises, and undoing it properly means deleting or re-merging the tail — a larger change than #131 asks for. Happy to file it separately.


rui-nar:
Filed the head-side follow-up as #141 — resetting the head restores the full Strava track while its tail rows survive, so the two overlap. Left out of this PR deliberately: the fix is a design call about what should happen to the tails, not a snapshot bug.

PR #142, opened 2026-07-28
fix(editor): undo the split when resetting its root, after confirming (#141)

### LINKED ISSUE #141: bug: "Reset to Strava" on a split head restores the full track while its tail still exists
### What

Resetting the **head** of a split restores the entire pre-split Strava track, but the tail row it was cut into is left alone. The two then overlap on the map: the head covers the whole original track, and the tail still covers its slice of it. The head's name stays `<name> (1/2)` even though it is no longer a half of anything.

This is the mirror of #131 (fixed in #140), which covered the tail side only. #140 deliberately left this alone because the head genuinely *is* the Strava activity, so "Reset to Strava" restoring the full track is arguably what the label promises — the question is what should happen to the tail, and that is a design call rather than a snapshot bug.

### Why

`split_activity` writes the head's geometry via `_write_track_geometry` (`src/project/repo_activities.py:361`), which snapshots the pre-split track into `original_polyline` / `original_elevation_profile_json` on the first edit (`src/project/repo_activities.py:112-114`). For a never-edited activity that snapshot is the full Strava track — correct in isolation.

`reset_activity_track` (`src/project/repo_activities.py:166`) then restores it and clears `is_edited`, touching **only that row**. It does not:

- delete or re-merge the tail (or tails — a family can be several pieces deep),
- call `_renumber_split_family` (`src/project/repo_activities.py:220`), so the surviving pieces keep their now-wrong `(i/N)` suffixes and the head keeps `split_base_name` set,
- clear `split_root_id` on the family members that pointed at this head.

### Repro

1. Open an activity's track editor, *Split here* at some point.
2. Reopen the head piece — `<name> (1/2)` — in the track editor.
3. Press *Reset to Strava*.
4. The head is back to the full original track, but `<name> (2/2)` is still in the timeline sitting on top of the head's second half. Both pieces keep their `(1/2)` / `(2/2)` names.

### Possible directions

- **Undo the whole split.** Reset on a split root deletes its local tails (reusing the local-delete path so items are unlinked and rows removed) and restores the plain base name from `split_base_name`. Closest to what a user pressing "reset" on a piece labelled `(1/2)` probably expects, but it silently destroys any edits made to the tails.
- **Confirm first.** Same as above, but the editor warns that resetting will remove the N pieces cut from this activity — so the destruction is a choice, not a surprise.
- **Reset only the head's own edits.** Treat the split as a boundary the way #140 now does for tails: snapshot the head's post-split geometry too, so reset undoes later edits but leaves the split standing. Consistent with the tail's behaviour, but then nothing in the UI can get back to the full Strava track, and the "Reset to Strava" label becomes wrong for a head as well.
- **Block it.** Refuse reset while the head still has surviving siblings, pointing the user at deleting the tail pieces first.

### Notes

Found while fixing #131. Pre-existing on `main`; #140 changes only the tail's snapshot and does not alter the head's behaviour either way.

rui-nar:
Fixes #141. Implements the **"confirm first"** direction from the issue. Follow-up to #140, now merged and rebased on top of.

## The bug

The root of a split family snapshots the whole pre-split Strava track — correct in isolation, it *is* the Strava activity. But `reset_activity_track` touched only that row, so restoring the full track left every piece cut out of it sitting on top of that track, and the root kept its now-meaningless `(1/N)` name.

## The fix

**Server** — reset on a family root undoes the split. `_undo_split` deletes each local piece, unlinks its timeline item, keeps item positions contiguous, and lets `_renumber_split_family` restore the plain pre-split name. Every piece of a family carries the *same* `split_root_id` (the first piece ever split, never the immediate parent), so one query reaches pieces cut out of other pieces too. Resetting a non-root piece cascades nothing — it still undoes only that piece's own post-split edits (#131).

The endpoint now passes the project id so items can be unlinked, taken from the already-resolved project row.

**Client** — that cascade destroys any edits on those pieces, so it can't be a surprise. The editor counts the pieces from the activity list and confirms first:

> **Reset to Strava**
> This restores the full original track from Strava, undoing the split that cut it into 2 pieces.
>
> The other piece will be removed from the trip, along with any track edits made to it. This cannot be undone.

This needs `split_root_id` client-side, so it's plumbed through `Activity` exactly the way `is_edited` already is — model field, `to_strava_dict`, `from_strava_api`, DB row mapper.

## Tests

Five API regressions in `tests/test_activity_split_api.py`; the three cascade ones fail with `_undo_split` disabled (verified):

- root reset removes the pieces, their timeline items and the `(i/N)` name
- it reaches pieces cut out of pieces (grandchildren)
- resetting a piece leaves its siblings alone
- an unsplit activity's reset touches no items
- `split_root_id` survives serialisation

Five widget tests in `flutter_client/test/activity_editor_page_test.dart` covering the warning text, cancel, confirm, and the two cases that must **not** warn.

All green on top of current `main`: **1214 Python**, **628 Flutter**, `flutter analyze` clean.

## Known remaining case

Resetting a piece that was *itself* split still overlaps its own children. Verified on this branch — split `[0..4]` at index 2, then split the tail at index 1:

```
-1: geom=[2.02, 2.03]        snap=[2.02, 2.03, 2.04]
-2: geom=[2.03, 2.04]
reset -1 →
-1: geom=[2.02, 2.03, 2.04]   # overlaps -2
```

`-1`'s snapshot predates the second split, and it has no pointer to `-2` — pieces point at the family *root*, not at their parent, so neither this PR's cascade nor #140's snapshot fix reaches it. Fixing it properly means adding a `split_parent_id` column plus a migration, which is a bigger change than #141 asks for. Happy to file it.


rui-nar:
Follow-up #145 (issue #143) is stacked on this branch — it replaces the `split_root_id`-keyed, root-only cascade added here with a transitive `split_parent_id` walk, which covers the re-split case this PR leaves open. Merging this first, then #145, works; #145 retargets to `main` automatically.

Issue #173, opened 2026-07-30
Write-path contention: item write granularity + durable job queue


rui-nar:
Infrastructure plan arising from #172 (concurrent segment route resolves). #172 stays the bug report with its tactical fix; this issue is the structural work behind it.

Concurrent resolves are one symptom of a broader shape: **every project mutation, however small, serializes at project granularity**, and **all background work shares the API process**. This covers both.

---

## Finding: there are two write paths to `projectitem`, and they race

**Whole-list rewrite** — `save_project` → `_replace_items` (`src/project/repo_items.py:22-64`) does `DELETE FROM projectitem WHERE project_id=?` then a single bulk INSERT of the entire list. 13 call sites across 6 files.

**Direct row insert** — `api/memories.py:335-341`, `api/journal.py:253-259`, `api/encounters.py:176-187` each shift `position += 1` on later rows and insert a `DBProjectItem` straight into the session. No `save_project`, no `lock_version` bump.

The good news: narrow item writes are already an established idiom in this codebase. Phase A extends it rather than inventing it.

The bad news: all three structural mutations in `api/project_items.py` — delete (`:55`), reorder (`:104`), sort (`:206`) — call `save_project` with `check_version` defaulting to `False`, i.e. a blind delete-all-and-reinsert with **no optimistic-lock check at all**. A memory/journal/encounter insert that commits after one of those loaded its snapshot is erased from the item list by the rewrite: the `memory` row survives, its linking `projectitem` row does not.

**Status: by inspection only — no repro written.** Treat as a lead. Confirming it is the first task of Phase A.

## The design principle

> **Structural writes** (add / remove / reorder) take the project CAS.
> **Payload writes** (resolved geometry, edited content) take a narrow row update and skip it.

A route resolve is a payload write. Under this rule it stops touching `lock_version` entirely, which means **Phase A alone closes defects 1 and 2 of #172 with no new infrastructure**. The broker then earns its keep on what granularity cannot fix: durability across restarts, and bounding Overpass concurrency (#172 defect 3).

---

## Phase A — item write granularity

**A1. Stable item identity.** Migration adding `uid` (uuid4 hex) to `projectitem`, unique on `(project_id, uid)`, backfilled. Uniform across all item types, not segment-only — reorder needs it too. `src/project/repo_transfer.py:161-168` also constructs item rows for import/transfer and must mint uids.

[...]

PR #174, opened 2026-07-30
Item write granularity: concurrent route resolves stop colliding (#173 phase A)

### LINKED ISSUE #172: Bug: concurrent segment route resolves collide — 409 on trigger, stuck `pending` on persist, Overpass rate-limit failures
Follow-up to #149.

Resolving more than one transport segment (ferry / bus / train) at the same time is supported by design, but three separate defects make it unreliable in practice. Each one is independent and has its own failure mode.

## Background — how it's meant to work

- `SegmentDialog._resolveAsync` is fire-and-forget (`unawaited(...)`, `flutter_client/lib/src/projects/segment_dialog.dart:416`) and the dialog closes immediately, so nothing prevents starting a second resolve.
- `POST /segments/{seg_id}/resolve-route` marks the segment `pending`, returns 202, and queues `_resolve_route_job` via `background_tasks.add_task` (`api/segments.py:431`).
- `_resolve_route_job` is a sync `def`, so Starlette runs it in the anyio worker threadpool — jobs genuinely run in parallel.
- Resolve state (`route_status`, `route_started_at`, `route_polyline`) is per-segment, so two segments never share status.

The concurrency model is sound. The problems are all in how the writes and the outbound Overpass calls behave under contention.

## Defect 1 — the trigger endpoint can 409, and the client doesn't recover

`resolve_segment_route` saves the project with `check_version=True` and **no retry** (`api/segments.py:428`). `save_project` does a compare-and-swap on `lock_version` (`src/project/repo_core.py:191-212`) and raises `StaleWriteError`, which the global handler maps to 409 (`api/router.py:153`).

Two triggers landing in the same project within the same instant — or a trigger racing a job that is just persisting its result — and the loser gets a 409. `ProjectService.resolveTrainRoute` does not retry 409 (`flutter_client/lib/src/projects/project_service.dart:264`), so it surfaces as an error toast and **that segment's resolve never starts at all**.

Note this is self-inflicted at load time too: `recoverStalePendingSegments` (`project_segment_crud_mixin.dart:347`) deliberately fires *N* resolves in a tight loop, which is exactly the collision pattern above.

**Symptom:** second ferry silently never resolves; user sees an error snackbar.

## Defect 2 — persist retry budget of 1 leaves segments stuck `pending`

The job's write loop is `for attempt in range(2)` (`api/segments.py:151-183`). Two concurrent resolves are fine — the loser catches `StaleWriteError`, reloads fresh, and re-applies. But **three or more** finishing near-simultaneously and one exhausts both attempts, hits `break`, and writes no terminal status.

The segment is left `route_status="pending"` with `route_started_at` set. The client polls for its 2-minute window (`pollSegmentResolution`, `project_segment_crud_mixin.dart:276`), times out, and shows "Still resolving route". Stale-recovery only re-triggers on the *next project load* after the 5-minute `_pendingStaleAfter` threshold.

**Symptom:** permanent spinner on the tile until the project is reopened. No data corruption — the whole-project save reloads before re-applying, so no other segment's result is clobbered.

## Defect 3 — Overpass rate limits multiply, and ferry/bus don't degrade

`_overpass` tries 3 mirrors, one attempt each, no backoff (`src/services/overpass_service.py:757`). Parallel resolves fire parallel Overpass queries from a single server IP, multiplying the per-IP rate.

Critically, **rail degrades to a straight endpoint chord on total failure, but ferry and bus raise `OverpassError` and go straight to `failed`** (`api/segments.py:64-66`). So N concurrent ferries are materially more likely to hard-fail than the same N run sequentially.

**Symptom:** ferry resolves that would succeed one at a time fail with an Overpass error when run together.

## Minor

Each job's `finally` does `bust_geo_cache` + `warm_geo_cache` over the whole project (`api/segments.py:193-198`), so N concurrent jobs re-warm the same cache N times. Wasteful but correct — the generation counter in `api/geo.py:52` prevents stale writes.

## Test gap

There is no test anywhere in `tests/` covering two resolves running at once. The existing concurrency tests cover tiles, tracks, DB pool and Strava (`test_tile_cache_concurrency.py`, `test_track_save_db_lock.py`, `test_db_pool.py`, `test_strava_rate_limit.py`) — none touch the resolve path.

---

Plan for the fix in the first comment. Execution deferred to a later session.

rui-nar:
Implements phase A of #173. Closes #172.

Full suite: **1527 passed, 0 failed.**

## The problem

`save_project` deleted every one of a project's `projectitem` rows and bulk-reinserted them, so item rows had no identity a targeted update could address. Every mutation, however small, serialised at project granularity behind one optimistic lock — with a retry budget of one. Three ferries resolving at once fought over that lock and a loser wrote no verdict at all, leaving a tile spinning on "pending" until the project was reopened five minutes later.

## The split

Writes are now separated by what they actually change:

| | goes through | takes the CAS |
|---|---|---|
| **structural** — add / remove / reorder | `save_project` | yes |
| **payload** — resolved geometry | single-row `UPDATE` | no |

A route resolve is a payload write, so resolves no longer contend with each other, and neither blocks an unrelated edit. That closes #172 defects 1 and 2 with no new infrastructure.

## What's in it

- **Migration `b5c8d2f1a930`** — adds `projectitem.uid` (durable per-item identity; the primary key churned on every save so it could never serve) and `projectitem.segment_id` (mirrored out of the JSON blob so it can be indexed). Both backfilled; the uid index is partial so any surviving NULL cannot collide under UNIQUE.
- **`_replace_items`** is now a uid-keyed diffed upsert — insert new, update changed, delete removed — instead of delete-all/reinsert.
- **Resolve trigger, job and crash safety net** are single-row writes, guarded by a `route_started_at` token so a job superseded by a newer trigger discards its stale verdict.
- **`save_project_with_retry`** replaces three inconsistent policies against the same conflict: none in the trigger (which surfaced a 409 the client never retried), a bare `range(2)` in the job, and none in `_mark_segment_failed` — so the crash safety net could itself lose the race and leave a segment pending.

## Also fixes a pre-existing data-loss race

This work surfaced a second bug with nothing to do with ferries. `delete`, `reorder` and `sort` all saved with **no version check at all**, while `api/memories.py`, `api/journal.py` and `api/encounters.py` insert item rows directly without advancing `lock_version`. A memory added concurrently with a reorder was silently erased from the timeline — the `memory` row surviving as an orphan with no timeline item.

Reproduced against unmodified code before any fix was written. Notably `lock_version` stayed at `0` through the direct insert, so `check_version=True` alone would **not** have caught it: the fix needed both halves. Direct inserts now advance the lock; the three structural endpoints take it.

## One design correction worth reading

The rule as originally written in #173 — "payload writes skip the CAS" — is wrong, and I caught it reviewing my own diff.

`save_project` rewrites **every item's payload** from the caller's in-memory snapshot. A write invisible to the lock is therefore not merely missed, it is actively *undone*: a resolve landing mid-reorder reverted the segment to "pending" and the resolved geometry was lost.

Corrected rule: a payload write **advances** `lock_version` without **taking** it. Resolves stay uncontended with each other; a concurrent structural edit now pays one retry instead of silently destroying the result. `test_a_structural_save_cannot_clobber_a_landed_resolve` covers it.

## Also included

`save_project` now refuses a project loaded with `journal_user_id`. Such a load hides other users' journal entries *and* their timeline items, so saving one would delete every filtered-out row. Not reachable today — all four callers are read paths — but it is one careless caller away from silent data loss, so it now raises instead.

## Notes for review

- **Every regression test was verified load-bearing** by neutering the fix and confirming the test fails: the direct-insert race, the clobber case, both resolve-contention tests, and the filtered-view guard.
- `test_migration_prune_orphaned_tails` needed a fix: it seeds ORM objects against a frozen historical schema, so any new `projectitem` column breaks it. The file already had a helper solving exactly this for `activity`; it is now generalised to `projectitem` too.
- **`low_res_geo_json` deliberately needs no refresh** on a narrow write — `api/geo.py:184` states the low-res endpoint always computes from the live project and never reads the cached column. Worth confirming you agree that column is genuinely write-only; nothing in `api/` or `src/` reads it.
- The `route_started_at` token guards *re-triggers*, not a segment edited and re-resolved into a coincidentally identical timestamp. A real job token belongs with the durable job rows in phase D.

Phases B–F of #173 (the Redis/RQ broker) are untouched.


rui-nar:
Closing — not abandoned. Phase A stays on `fix/173-item-write-granularity`; phases B–F of #173 will land on the same branch, and a single PR will be opened once the whole issue is done.

PR #182, opened 2026-07-30
Write-path contention: item granularity + durable job queue (#173, all phases)

### LINKED ISSUE #172: Bug: concurrent segment route resolves collide — 409 on trigger, stuck `pending` on persist, Overpass rate-limit failures

### LINKED ISSUE #173: Write-path contention: item write granularity + durable job queue

rui-nar:
Implements all of #173 (phases A–F). Closes #172, closes #173.

**Python: 1558 passed. Flutter: 725 passed.** 44 files, +2601/−331.

Supersedes #174, which was phase A only and was closed so this could land as one reviewable change.

---

## Why

Two structural problems, one symptom. `save_project` deleted every one of a project's `projectitem` rows and bulk-reinserted them, so item rows had no identity a targeted update could address — every mutation serialised at project granularity behind one optimistic lock with a retry budget of one. And all background work ran as `BackgroundTasks` in the API process, so a restart lost it and nothing bounded outbound concurrency.

Three ferries resolving at once hit both: they fought over the lock, a loser wrote no verdict, and the tile spun on "pending" until someone reopened the project five minutes later.

## The rule this establishes

> **Structural writes** (add / remove / reorder) take the project CAS.
> **Payload writes** (resolved geometry) take a narrow row update and skip it.

A resolve is a payload write. That alone closes #172 with no new infrastructure.

**One correction to the plan as published in #173.** "Payload writes skip the CAS" loses data. `save_project` rewrites every item's payload from the caller's snapshot, so a write invisible to the lock isn't merely missed — it's actively *undone*: a resolve landing mid-reorder reverted to "pending". The working rule is that a payload write **advances** `lock_version` without **taking** it. Resolves stay uncontended with each other; a concurrent structural edit pays one retry. Caught while reviewing the diff, reproduced, fixed, and covered by `test_a_structural_save_cannot_clobber_a_landed_resolve`.

## By phase

**A — item write granularity.** Migration `b5c8d2f1a930` adds `projectitem.uid` (durable identity; the primary key churned on every save so it could never serve) and `segment_id` (mirrored out of the JSON so it can be indexed). `_replace_items` becomes a uid-keyed diffed upsert. One `save_project_with_retry` replaces three inconsistent policies against the same conflict — none in the trigger (a 409 the client never retried), a bare `range(2)` in the job, none in the crash safety net, which could therefore lose the race it existed to catch.

**Also fixes a pre-existing data-loss race this surfaced.** `delete`, `reorder` and `sort` saved with *no version check at all*, while memories/journal/encounters insert item rows directly without advancing `lock_version`. A memory added concurrently with a reorder was silently erased from the timeline, its row orphaned. Reproduced against unmodified code first — and notably `lock_version` stayed at 0 through the direct insert, so `check_version=True` alone would not have caught it. Both halves were needed.

**B — prerequisites for out-of-process work.**
- **B1 was a latent bug, not an optimisation.** The geo cache is a process-local dict, so a worker's `bust_geo_cache` would drop only its own copy while the API served pre-resolve geometry for 300 s. The generation counter moves to Redis and each entry records what it was computed against.
- **B2 needed no code change** — WAL, `busy_timeout=30000` and `synchronous=NORMAL` were already per-connection. But nothing exercised *two engines on one file*, so it does now. Only the API checkpoints the WAL, which is correct: one file, one WAL, and its checkpoint folds in the worker's frames.
- **B3** `VIEWTRIP_ROLE` gates migrations, admin seed and scheduler to the API process. Two containers racing `alembic upgrade head`, or each taking its own nightly backup and 60 s WAL checkpoint, is silent and periodic.
- **B4** `/metrics` aggregates via `PROMETHEUS_MULTIPROC_DIR` when set. The docs that asserted one process saw everything are corrected — unset, a scrape *under-reports rather than fails*, which is the trap worth knowing.

**C — Redis + RQ.** An RQ worker runs one job at a time, so the resolve queue's worker count *is* the Overpass bound — enforced regardless of how many API processes exist, which the `threading` semaphore originally proposed in #172 could never do (it silently stops bounding anything at `--workers 2`).

**D — durable job rows.** `DBRouteJob` mirrors `DBPosterJob`. The row, not the queue entry, is the record that work is owed, because every layer above it can lose the job. A startup sweep re-queues anything non-terminal — deliberately *not* age-scoped, since at startup nothing is executing. That is precisely what the client couldn't know, which is why its version needed a five-minute window. A poison job is abandoned after 3 attempts and its segment failed, so the tile stops spinning rather than promising a result that will never come.

**E — remaining background work.** Poster, share tiles and stats move onto the queue.

**F — deletes `recoverStalePendingSegments`.** Recovery from a *server* crash was implemented in the Flutter client. It's the server's job now, so the client code is removed rather than fixed.

## Redis is optional throughout

With `REDIS_URL` unset every job runs in-process exactly as before. That is the supported self-hosted setup, what the test suite uses (via `fakeredis`, so CI needs no broker), and the rollback lever — unset it, no image rebuild. A configured-but-unreachable broker degrades the same way rather than failing requests.

## Deploying this

`deploy.ps1` needs **no changes** — it builds/pushes the image then runs `docker compose down && pull && up -d`, which names no services. The whole deployment change is:

1. Add the `redis` + `worker` services to the compose file on val and prod (see `docker-compose.yml.example`).
2. Add `REDIS_URL` (and `PROMETHEUS_MULTIPROC_DIR`, if you scrape metrics) to each host's `.env`.

Keep the worker on the **same image tag** as the API: they share one image and only the API runs migrations, so a worker on an older tag would run stale job code against a schema it does not know about.

Needs a version bump. I have not run any deploy.

## Notes for review

- **Every regression test was verified load-bearing** by neutering its fix and confirming failure: the direct-insert race, the clobber case, both resolve-contention tests, the filtered-view guard, and cross-process cache invalidation.
- **`docker-compose.yml.example` is new.** The real compose is gitignored for holding secrets, so there was nowhere the worker/redis wiring could be recorded. It flags the three easy mistakes: the worker must mount the same host paths (SQLite is a file), Redis needs `--appendonly yes` or a broker bounce silently drops every queued job, and Redis' port must not be published (Docker's DNAT rules bypass `ufw`).
- **`save_project` now refuses a project loaded with `journal_user_id`.** Such a load hides other users' journal entries *and* their timeline items, so saving one would delete every filtered-out row. Not reachable today — all four callers are read paths — but one careless caller away from silent data loss.
- **Still on `BackgroundTasks`:** photo downloads and outbound email. Out of scope for #173 and neither leaves a user-visible object stuck pending the way a lost resolve did.
- The `route_started_at` token guards *re-triggers*, not a segment edited and re-resolved into a coincidentally identical timestamp. `DBRouteJob.id` is now available to become a real token if that ever matters.


rui-nar:
**Correction to the "Deployment changes" bullet above: `deploy.ps1` needs no changes.**

I asserted it did without checking. It builds/pushes the image and then runs `docker compose down && pull && up -d && ps` over SSH — no container names, no health checks, no per-service logic. New compose services are picked up automatically, and `pull` fetches `redis:7-alpine` alongside the app image.

The deployment change is exactly:

1. Add the `redis` + `worker` services to the compose file on val and prod (see `docker-compose.yml.example`).
2. Add `REDIS_URL` (and `PROMETHEUS_MULTIPROC_DIR`, if you scrape metrics) to each host's `.env`.

Nothing else. Fixed in 2623299, along with one thing worth recording that this raised: the worker must stay on the **same image tag** as the API. They share one image and only the API runs migrations, so a worker left on an older tag would run stale job code against a schema it does not know about.

The version bump still applies.

Issue #292, opened 2026-09-02
perf(map): Phase 1 — decode heavy payloads off the UI isolate


rui-nar:
The largest remaining main-thread block during a cold trip open. See [docs/PERF_MAP_LOAD.md](../blob/main/docs/PERF_MAP_LOAD.md).

`ApiClient._handle` does a bare `jsonDecode(res.body)` for **every** endpoint, including the ~12 MB details payload and the full-res geo. `ProjectService.getGeo` then runs `expandEncodedActivities` — a Google-polyline decode over potentially 300k+ points — inline immediately after.

Earlier rounds moved the *derived* computations (`buildFullTrackResult`, `hitTestMapTap`, `computeElevationSpots`, `decimatePolylinePoints`) behind `compute()`. Those are the cheap half. The parse that produces their input has no threshold guard at all.

Telling detail: a **warm** open is smoother than a cold one, because `project_cache_store_native.dart` already gunzips and decodes via `compute()`. The disk path is protected; the network path is not.

**Scope**
- Fetch heavy payloads as bytes (`bodyBytes`), not `String`.
- One isolate hop doing `utf8.decode` -> `jsonDecode` -> domain transform, with `expandEncodedActivities` folded into it rather than run separately on the UI isolate.
- Route `getGeo`, `getDetails`, `getLowResGeo` through it.
- Error bodies stay on the string path (they are small, and `ApiException` needs the text).

**Non-goals**
- Web still runs `compute()` inline; this phase must not regress web, but fixing it is Phase 4's payload reduction, not this.

**Verify**
- `decode_geo` / `decode_details` spans (#291) no longer appear on the UI thread.
- #276's repro — pan immediately after the spinner clears — is smooth on Android.

Part of #276.

Issue #276, opened 2026-08-27
ANR on post-load 


rui-nar:
Executing a map panning action within the first seconds after a post-load "seems" complete triggers an ANR. 
The user needs to wait a lot of seconds before doing anything not to block the app. 

After careful check, I believe the blocker is the elevation chart. I can see the elevation curve going from low-res to high-res and after that I can move around without an ANR.

rui-nar:
Audited the whole load path (client fetch -> decode -> notifier -> map rebuild, plus the server geo/tile endpoints). Write-up in `docs/PERF_MAP_LOAD.md`; staged plan broken out into #291-#296.

Short version of the diagnosis: the ANR fixes so far moved the *derived* computations off the UI isolate (`buildFullTrackResult`, `hitTestMapTap`, `computeElevationSpots`, `decimatePolylinePoints`) but never the **decodes that produce their input**. `jsonDecode` of the ~12 MB details payload and of the full-res geo, plus `expandEncodedActivities`' polyline decode, all still run on the UI isolate — that is the largest unguarded main-thread work left in the app.

The "small local blocks and unblocks" specifically is `_loadFullGeoProgressively`'s 8-batch reveal: it stages 8 full-tree rebuilds 80 ms apart for a payload that has already fully arrived and been parsed. The symptom has a metronome shape because the cause is one.

Your own note here — "I see the elevation curve go low-res to high-res and after that I can move around" — is the 12 MB `getDetails` parse. That is #292, and #295.2 deletes the payload entirely.

Starting on #291 + #292.

rui-nar:
The reported symptom is fixed. Closing this as the umbrella; the remaining work has its own issues.

**Root cause of the freeze**: `compute()` copies its *argument*, and the calling isolate does the copying (results are not copied — `Isolate.run` returns via `Isolate.exit`). Two per-trip computations were handed the whole `activities` list, whose `elevation_profile` is one Dart list object per sample — ~1.47 M objects. Both were serialising a million-odd objects **on the UI isolate in order to move work off the UI isolate**. That was the 2437 ms frame.

Three earlier hypotheses were falsified by measurement first, GC over a large heap foremost. What finally located it was the worst-frame context line: a 2437 ms frame whose only named span measured 0.3 ms.

**Measured, first report to last:**

| | before | after |
|---|---|---|
| worst frame | 2437 ms | **176 ms** |
| frames > 500 ms | 1 | **0** |
[...]

PR #298, opened 2026-09-02
perf(map): Phases 2-3 — one geo swap, geometry derived off the UI isolate, cheap selection

### LINKED ISSUE #293: perf(map): Phase 2 — typed geometry model, drop the progressive geo reveal
The targeted fix for the "small local blocks and unblocks" symptom in #276. See [docs/PERF_MAP_LOAD.md](../blob/main/docs/PERF_MAP_LOAD.md).

Two coupled problems:

**The progressive reveal is now pure cost.** `_loadFullGeoProgressively` stages ~8 geo swaps 80 ms apart. That made sense when the upgrade genuinely trickled in; today the whole payload has already arrived and been parsed before the first batch fires. It buys no perceived progress and costs 8 full-tree rebuilds (every polyline spec, every marker spec, a `compute()` decimation hop, `ActivityPanel`, `ElevationChart`). The symptom has a metronome shape because the cause literally is one.

**Geometry is untyped GeoJSON, re-walked on every rebuild.** `_buildPolylineSpecs`, `_buildActivityMarkerSpecs` and `buildFullTrackResult` each walk `f['geometry']['coordinates']` with `as num` casts per coordinate and allocate a `LatLng` per point. Rebuild is O(total points), not O(activities).

**Scope**
- `TrackGeometry`: per activity, `Float64List` lat/lon + precomputed bounds, built once inside Phase 1's (#292) isolate hop.
- Spec builders consume it.
- Replace the 8-batch reveal with a single atomic swap scheduled at `Priority.idle` via `SchedulerBinding.scheduleTask`.
- Replace `_waitForCameraIdle`'s 100 ms poll loop with that scheduler priority — it currently only defers the stall, and gives up after 2 s, landing the full rebuild mid-gesture anyway.

**Verify**
- A test counting `notifyListeners()` during a full load: exactly one geometry-driven rebuild.
- Spec-build spans (#291) drop to O(activities).

Depends on #292. Part of #276.

rui-nar:
Stacked on #297 (base is `perf/276-map-load-pipeline`, not `main` — merge that first).

This is the one that targets the symptom in #276 directly: *"still some jerkiness, some small local blocks and unblocks."*

## What it does

`_loadFullGeoProgressively` replaced the low-res track a few activities at a time, notifying after each batch — ~8 repaints 80 ms apart. Every one rebuilds the whole tree: each polyline spec, each marker spec, a `compute()` decimation hop, `ActivityPanel`, `ElevationChart`.

The staging made sense when the upgrade genuinely trickled in. It does not now — `getGeo` returns, and since #292 fully *parses*, the entire payload before the first batch could fire. The authoritative "final pass" the loop ran afterwards was already a complete atomic swap, so deleting the loop leaves correct behaviour with **one** rebuild instead of nine.

`_waitForCameraIdle` is now event-driven off `setMapCameraActive` instead of polling every 100 ms for up to 2 s — that polling was itself work competing with the gesture it was trying to stay out of the way of. Same cap, now injectable so tests don't wait real seconds.

## Measured

The new test asserts what the UI actually *observes*: across a whole load, listeners may be handed at most two distinct `geo` objects — the low-res one, then the full-res one. Run against the batched implementation first:

```
Expected: an object with length of <2>
listeners saw 9 distinct geo objects; the upgrade must be
low-res then one atomic full-res swap
```

**9 → 2.**

## A plan change worth flagging

The plan in `docs/PERF_MAP_LOAD.md` said to schedule the swap via `SchedulerBinding.scheduleTask(..., Priority.idle)`. I probed that before building on it, and it does not hold up:

- an idle task resolves in ~3 ms in a plain `test()`, but **does not run at all under a pumped `testWidgets` pipeline** — gating the upgrade on one hangs the background load in widget tests;
- the same objection applies in production: a busy map is precisely the situation with no spare frame time, so the upgrade would be starved exactly when it is most wanted.

Rejected, with the reasoning recorded in the doc so it isn't retried blind. The event-driven camera-idle wait gets the intended benefit without the frame-pipeline dependency.

## Housekeeping

- `progressiveGeoBatchSize` and its test are deleted along with the loop they existed for.
- `progressive_geo_cache_shortcut_test` keeps its warm-cache invariant (independent of the reveal, still load-bearing); only its now-stale narration changed.

Full suite: **1180 passing**. `flutter analyze`: clean.

Fixes #293. Refs #276.


rui-nar:
Added **Phase 2b** to this PR rather than stacking a third branch — same issue (#293), and a stacked PR gets no CI until it targets `main`.

**2b did not go as planned, and the plan was wrong rather than the code.**

The phase called for a `TrackGeometry` model holding `Float64List` lat/lon per activity, premised on "every rebuild re-walks GeoJSON coordinates, so rebuild cost is O(total points)". I benchmarked that premise before refactoring and it does not hold: `memoCoordsToLatLng` already memoizes the conversion on each feature's coords-list identity. On a 100-activity / 500k-point trip:

```
cold=83.4ms   warm(each)=0.114ms
```

So rebuilds were never O(points) — only the *first* build after a geo swap is. And `flutter_map`'s `Polyline` takes a `List<LatLng>`, so typed buffers would have to be materialised back into exactly that list on the render path: cost moved, not removed. The full refactor would have touched the spec builders, `buildFullTrackResult`, `hitTestMapTap`, `extractSelectedPoints`, the segment-overlay merge and the on-device cache — a wide blast radius for no measured gain. Rejected, with the reasoning and the numbers recorded in `docs/PERF_MAP_LOAD.md`, including a correction to the root-cause section where I originally got this wrong.

**What shipped is the part of the idea that survives the measurement.** The decode hop from #292 now produces the coordinate→LatLng conversion alongside the parse and seeds the cache with it. Both halves ride back on one zero-copy hop, and since the coords lists and their converted points transfer as a single object graph, identity-keyed caching still works across the isolate boundary. The UI isolate never pays the ~83 ms cold conversion at all.

That conversion was the last O(points) cost left on the UI isolate in the geo path once 2a cut nine swaps down to one.

**Verified by counting, not timing:** after `decodeGeoOffIsolate`, walking every feature through `memoCoordsToLatLng` performs **zero** conversions — with a baseline test showing an unseeded geo pays one per feature, so the assertion can't go vacuous.

Full suite: **1184 passing**. `flutter analyze`: clean.

PR #343, opened 2026-09-06
perf(server): cut Overpass demand, and back off instead of retrying

rui-nar:
Load reduction. Follows #341 and #342.

## Why this became urgent

Today's investigation ended with this deployment's **IPv4 address refused outright**
by both overpass-api.de backends:

```
gall     (162.55.144.139)  connection refused  0.008s
lambert  (65.109.112.52)   connection refused  0.027s
example.com over IPv4      200                 0.086s
overpass-api.de over IPv6  200                 1.383s
```

TCP RST in milliseconds, while IPv4 elsewhere and IPv6 to the same host both work.
The escalation is legible in one worker log — soft 429s from 15:36, a 504 at
15:37:11, first connection failure at **15:37:53**, nothing since. We earned it.

**This PR does not restore service.** The block has to lift, or the containers need
a working IPv6 path. It is what stops us re-earning the block on whichever address
we end up using — which is why it comes first.

## Concurrency: 2 → 1

`_OVERPASS_CONCURRENCY` matched the advertised per-IP cap exactly, which leaves no
margin. Two resolve workers each holding a query sat permanently on the boundary,
and Overpass doesn't free a slot the moment our socket closes.

The cost of being wrong here was worse than I assumed when writing #341. **A 429 is
not free** — the dispatcher queues you *before* refusing:

```
busy after 11.6s   busy after 12.3s   busy after 12.8s   busy after 14.1s
```

One measured resolve spent **~48 s of its ~130 s** simply being told to go away.
A healthy query answers in 1.5–2.6 s, so serialising our own requests is far
cheaper than contending for the last slot.

## Caching

Every Overpass query is a pure function of the segment's coordinates, so a retry —
by hand, or by `sweep_degraded_segments` — re-asks byte-identical questions. Nothing
cached them.

`src.jobs.upstream_cache` stores bodies in Redis, gzipped (OSM geometry is mostly
repeated coordinate digits, ~10× compression) under a ceiling on the **compressed**
size — that broker also holds the RQ queues inside 256 MB, and judging entries on
raw size would refuse exactly the ones worth keeping.

- Checked **before** the pacing gate: a hit costs no slot and no wait.
- **Only parsed successes are stored.** Caching a 429 or a timeout would turn a
  transient upstream problem into a day-long one.
- Best-effort throughout: a cache that cannot answer is a miss, never an error.
- 24 h TTL — covers every repeat we actually make (a person retrying within
  minutes, the sweep's five hourly attempts) without pinning a wrong answer for a
  week after someone fixes the underlying OSM data.

## Tests

`tests/test_upstream_cache.py` (10) and `TestResponseCaching` / `TestPolitenessBound`
in `tests/test_overpass_failover.py`. Both new integration groups were verified to
**fail with the cache stubbed out** and pass with it:

```
FAILED …::test_an_identical_query_is_not_re_requested
FAILED …::test_a_cache_hit_costs_no_slot_and_no_wait
```

Full suite: **2370 passed**.

## A correction to #341's rationale

#341's endpoint measurements were taken with bare `curl`, which prefers IPv6 on a
dual-stack host — while the workers run in containers that have no IPv6 at all.
So "overpass-api.de answers a full 5.2 MB query in 8.4 s" described a path the
application cannot use. The conclusion drawn from it — that the upstream was
healthy and the problem entirely self-inflicted — was half right: the load was
ours, but the path was also broken.

What survives unchanged is everything measured from inside the containers: the
429s are real, and waiting them out demonstrably works. What does not survive is
private.coffee's "100 % failure rate", measured over the same broken IPv4 — so it
is **not** retired here, and the circuit breaker is deliberately still deferred
until it can be judged on a working network path.


rui-nar:
## Second commit folds in the documentation review

A research pass over the OSM wiki, `commons.html`, the Overpass server source and the operators' forum posts found that **the retry behaviour in the first commit is the named trigger for the ban we received**, so this branch now also rewrites it.

### The finding

Policy for overpass-api.de: *"If you receive an HTTP error code such as 429 or 406, **pause for 30 seconds** before making a new request."* Operators have added that clients repeatedly hitting a 429 are banned faster, and that a high 429 *rate* is a trigger by itself.

We waited **5 seconds** and retried the same host three times.

Both paths meant to derive a real wait were **inert**: Overpass sends no `Retry-After` at all, and `"in N seconds"` appears only in `/api/status`, never in a 429 body. Production logs confirm the 5 s default is what ran every time — every line reads `waiting 5s`, never a parsed value.

There is also a forum thread from nine days ago reporting the same symptom on the **same two backend IPs**, answered by an OSM sysadmin: *"the IP had been blocked for ignoring the 429 return."*

### What replaces it

No in-band retry. A back-off status marks the **host** as cooling and the query moves on or degrades — the host being the right unit, since a resolve makes several queries and pausing one while the next hits the same host misses the point. Cooldowns are shared through Redis so one back-off is observed by every worker.

| trigger | cooldown | why |
|---|---|---|
| 429 / 502 / 503 / 504 | 60 s | documented 30 s minimum, with margin. 504 is included on the server's own meaning — resource admission refused, not a generic gateway error |
| connection failure | 3600 s | how a block presents; an operator asked whether continued attempts prolong one could not rule it out |

### Also corrected

- **User-Agent** pointed at `github.com/viewtrip`, which doesn't exist. UA bans are **manual and don't expire**, unlike IP bans — an unverifiable client invites the permanent kind. Now the real repo and the real `APP_VERSION`.
- **Socket timeout 45 s → 60 s.** The server enqueues up to 15 s before admitting, then runs up to the declared 30 s; 45 s sat exactly on the worst compliant case and aborted valid requests, burning the slot for nothing.
- **"Rate limit: N" is not concurrency.** It counts slots held for execution time *plus* a load-dependent cooldown. Comment corrected; concurrency stays at 1.
- **kumi.systems *is* private.coffee** renamed — the endpoint list was one operator counted twice.

Full suite: **2372 passed**.

### Not in this branch

**Volume.** Fair use is ~100 requests and ~10 MB/day *summed across all users of an application*. One bbox query returns 5.2 MB. No client-side politeness fixes that — it needs a local data source.

Issue #345, opened 2026-09-06
Replace Overpass with a locally-held rail extract


rui-nar:
Route resolution depends on a free public service that has now blocked this deployment's IPv4 address. A spike shows the data we actually need is small enough to hold locally, which removes the dependency entirely rather than negotiating with it.

## Why now

On 2026-09-06 overpass-api.de began refusing TCP connections from our VPS on both backends (`162.55.144.139`, `65.109.112.52`) — instant RST, while IPv4 to other hosts and IPv6 to the same host both work. Escalation was soft 429s → 504 → hard refusal over about two minutes.

A documentation review (wiki, `commons.html`, server source, operator forum posts) found:

- The stated policy is **pause 30 s** after a 429; we waited 5 s and retried. That is the named ban trigger, and a forum thread from nine days earlier reports the same symptom on the same two IPs, answered by an OSM sysadmin: *"the IP had been blocked for ignoring the 429 return."*
- Fair use for the main instance is roughly **100 requests and 10 MB per day, summed across all of an application's users**. **One of our bbox queries returns 5.2 MB.**
- *"Setting up an app for more than just OSM mappers and relying on the public instances as backend… only running your own instance sustainably serves your mission."*

#341, #342 and #343 fix our client behaviour. **None of them fix the volume**, which is the part that cannot be tuned away.

## Spike results — measured, not estimated

Filter = the exact tag selection our Overpass queries already use: `railway in (rail, narrow_gauge, light_rail)` without `service`, `route=train` relations, stations with `uic_ref`.

| | raw | rail-only PBF | graph file | ways | uic stations | filter time |
|---|---|---|---|---|---|---|
| Denmark | 494 MB | **0.31 MB** (1,596×) | 1.2 MB | 5,392 | 2 | 197 s |
| Germany | 4.83 GB | **10.0 MB** (483×) | 26.1 MB | 130,713 | 5,483 | 2,022 s |
| Europe | 34.9 GB | *(projected ~50–90 MB)* | *(~150–250 MB)* | | | |

**Europe raw does not fit on the current 40 GB VPS. Europe rail does, comfortably, with no tier upgrade** — provided the filtering happens off-box.

### Performance, using our existing `_build_rail_graph` / `_nearest_node` / `_dijkstra`

Germany (1,137,563 nodes), as-is with no optimisation:

```
pickle load        0.19 s
graph build        8.41 s        319 MB resident
[...]

PR #347, opened 2026-09-06
feat(ci): build rail-only OSM extracts in CI (#345, phase 1)

rui-nar:
## Summary

Phase 1 of `docs/LOCAL_RAIL_DATA_PLAN.md` (issue #345): the CI pipeline that turns Geofabrik country extracts into rail-only artifacts, so the 34.9 GB Europe extract never has to reach a 40 GB VPS.

Nothing outside the pipeline is touched — no resolver, no store, no client. The output is exactly the phase 1 / phase 2 contract: per-region `<slug>-rail.osm.pbf` plus a `manifest.json` in the schema the plan fixes.

**What's here**

- `scripts/build_rail_extract.py` — filters one region. `osmium tags-filter` (C++) does the pass over the raw file; a pyosmium pass over *its* output applies the exact selection. The split exists because `tags-filter` can only OR tag patterns together — it cannot say "railway ways *without* a service tag" or "station *and* uic_ref" — so it can only over-select, and the exact pass is affordable only because it runs on 1 MB instead of 4.83 GB.
- `config/rail_regions.yml` — 49 European countries. The workflow reads its build matrix from this file at run time, so adding a country is an edit here plus a rebuild, never a code change (plan, phase 0).
- `.github/workflows/rail-extract.yml` — monthly + `workflow_dispatch` (with an optional region subset for re-running one country). Matrix over the config, then a publish job that verifies checksums and uploads to a `rail-data-<date>` pre-release.
- Tests: the selection pinned against a checked-in 420 KB fixture, the manifest schema and its checksum verification, and guards that no raw extract can reach the repo or an image.

## The selection, after #349

The first version of this PR implemented the plan's original filter, which was narrower than the queries in `src/services/overpass_service.py`. #349 corrected the plan; the second commit here implements the corrected table, verified row by row against that source file rather than against the plan.

| | selection | Overpass query it mirrors |
|---|---|---|
| ways | `railway` in (`rail`, `narrow_gauge`, `light_rail`), no `service` tag | `_via_coordinate_fallback` |
| relations | `route` in (`train`, `railway`, `light_rail`) | `_route_relation_segment`, `_via_train_relations_endpoints` |
| nodes | **any** node carrying `uic_ref` | `_route_relation_segment` |
| stations | node, way **or relation** with `railway` in (`station`, `halt`) and `uic_ref` | `_find_station_near` |

Keeping a station that is a way or a relation means keeping what gives it a position — `out center` needs geometry — so `select` gained a pass for a station relation's member ways and their nodes.

**Recounted, against fresh 2026-09-06 extracts.** The earlier figures were cut from 2026-09-05 files, so a few counts move by OSM drift rather than by selection; that is the `ways` column.

| | ways | relations | stations | uic nodes | size |
|---|---|---|---|---|---|
| Denmark, narrow | 5,392 | 93 | 2 | — | 0.63 MB |
| **Denmark, corrected** | **5,394** | **196** | **2** | **3** | **0.67 MB** |
| Germany, narrow | 130,713 | 2,179 | 5,483 | — | 19.98 MB |
| **Germany, corrected** | **130,729** | **3,111** | **5,511** | **23,577** | **21.28 MB** |

What moved and why:

- **ways** +2 / +16 — drift. The selection for this row did not change.
- **relations** +103 / +932 — `route=railway` and `route=light_rail`. Denmark's 93 train relations sit beside 76 railway and 27 light_rail ones; that is two thirds of the country's route relations previously discarded.
- **stations** +0 / +28 — station ways and relations. Germany has 8 station ways and 20 station relations carrying a `uic_ref`; Denmark has none, so its count is unchanged and its 2 remain the two the spike found.
- **uic nodes** — the row that matters. Germany holds **23,577** nodes with a `uic_ref` against 5,483 tagged as stations: **four fifths of the lookup surface** was being dropped, exactly as #349 describes. Denmark holds 3 (two stations and a `railway=border` node) — its `uic_ref` coverage is genuinely that thin.

Germany's `bbox` is `[5.935, 47.397, 15.062, 54.922]` — the true extent, and it reaches *past* the nominal east edge because Geofabrik's cut overlaps the border.

## The member-way closure

The corrected table left one gap, which the third commit closes. Overpass answers a relation query with `out geom` — geometry for **every** member. Returning only the members that pass the way row is a behaviour change dressed as a filtering detail: `_extract_relation_geometry` builds a graph from the members it has and returns `None` when that graph is disconnected, so strategies A and B fall through where Overpass succeeds, looking exactly like "no route found".

So a kept relation's member ways are now held whatever their own tags, with their nodes.

| | route relations | missing ≥1 member, before | after |
|---|---|---|---|
| Denmark | 196 | **171** | **22** |
| Germany | 3,111 | **2,942** | **399** |

**Nothing new marks the members as unroutable, deliberately.** `is_rail_way` is the whole distinction, and phase 2's builder already applies that exact predicate to set the `rail` flag that keeps a way out of its R-tree and out of snapping — `int(tags.get("railway") in _RAIL_VALUES and "service" not in tags)`, character for character my `is_rail_way`. Inventing a second marker would have given the two phases two ways to disagree. A platform the resolver can route over would be worse than a missing one, so the tests assert both halves: the members are present, and none of them is rail.

**Size cost — much smaller than the count suggests, as you predicted.**

| | ways | member ways | size before | size after |
|---|---|---|---|---|
| Denmark | 5,394 | +1,889 | 0.67 MB | **0.80 MB** (+19 %) |
| Germany | 130,729 | +28,993 | 21.28 MB | **24.98 MB** (+17 %) |

A 35 % rise in Denmark's way count for 19 % more bytes: the members are platforms and sidings, and they are short. Europe lands around 130–220 MB.

**Residual cross-border gap, for phase 3.** Counted by member slot, which is what reconstructing a relation actually needs:

| | slots | unresolved before | after |
|---|---|---|---|
| Denmark | 24,548 | 27 % | **16 %** |
| Germany | 736,417 | 26 % | **19 %** |

The closure removed everything it could; what is left is ways the country extract does not contain. I checked that rather than assuming it: 20 of Germany's unresolved members, sampled at random, run through `osmium getid` against the raw 4.5 GB extract — **none of the 20 is in it**. Germany's 19 % is higher than Denmark's 16 % because its long-distance relations run to Basel, Vienna and Amsterdam. `member_ways_missing` is in the build log per region, so phase 3 can size the case before it writes the cross-border logic.

## The published fixture for phase 2

`tests/fixtures/rail_mannheim_filtered.osm.pbf` (174 KB) is the fixture put through the filter — a `.pbf` phase 2 can build a store from directly, without invoking this script. A test regenerates it and compares **byte for byte**, so changing the selection without regenerating it fails CI.

That is the anti-drift mechanism the review of #348 showed was missing: phase 2's own fixture was hand-cut with the pre-#349 filter, which left its `relation_uic` table empty and strategy A untestable against real data — and nothing could see it. This file also carries what that fixture lacked: 11 uic-bearing stop nodes, a station mapped as a way, a station mapped as a relation, and seven member-only ways (four sidings, three platforms).

Per the coordination note, #348 finishes against its own fixture and adopts this one as a follow-up; this PR is not held open for it.

## Do the tests actually detect a wrong selection?

Widening a filter can make an assertion vacuously true, so each row was mutated and the suite re-run. All eight go red:

| mutation | detected |
|---|---|
| relations narrowed to `route=train` (the pre-#349 contract) | 4 failures |
| nodes narrowed to station/halt + `uic_ref` (the pre-#349 contract) | 6 failures |
| station **ways** dropped | 4 failures |
| station **relations** dropped | 2 failures |
| `service` ways no longer excluded | 5 failures |
| way types widened with `tram` | 3 failures |
| way types narrowed to `railway=rail` | 4 failures |
| geometry closure dropped (kept ways lose their nodes) | 1 failure |

Beyond the pinned counts, four "the fixture is worth testing against" tests assert that each row is actually *represented* in the fixture — a uic node that is not a station, a station mapped as a way, more than one route type, service ways to exclude — so a future fixture that is too thin fails rather than passing vacuously.

The fixture was re-cut for this: the old Aarhus box had no uic-bearing node that was not a station and no station mapped as a way or relation, so it could not have caught any of the #349 breakage. It is now two 900 m boxes in Mannheim merged with `osmium merge` — the ARENA/Maimarkt halt (a station way) and Neuostheim (a station relation), 11 uic stop nodes between them, 27 narrow-gauge and 12 rail ways to keep against 152 service ways, 49 trams, 12 platforms and a signal box to drop, plus `route=train` and `route=railway` relations against buses, trams, cycle routes, a pipeline and a waterway. 420 KB, from two boxes rather than one because the dense kilometre between them costs 600 KB and decides nothing.

Three more tests cover the closure specifically: the members are kept whatever their tags (a service siding and a platform are asserted by name), none of the member-only ways is rail, and no service way is in the file except as a member. The fixture was chosen so those cannot pass vacuously — it holds four sidings and three platforms pulled in by relations.

## Two more things worth reading

**1. The artifacts are ~2.5× the size quoted in #345** (0.80 MB vs 0.31 MB for Denmark, 25.0 MB vs 10.0 MB for Germany). Roughly half of that gap is the member-way closure above, which #345's figure predates; the rest is encoding, not selection. Identical element and node counts to the spike's, and `osmium cat` reproduces this writer's output to within 3 bytes; even a variant with *every tag removed* is 0.53 MB for Denmark, so 0.31 MB was never reachable with tags intact. Europe lands around 130–220 MB rather than 50–90 MB, still a rounding error on a 40 GB box and on a release asset.

**2. One dependency added**: `osmium>=4.0.0` in `requirements.txt`, marked CI-only alongside the existing `PyYAML` precedent. A ~2 MB wheel that nothing at runtime imports, but it does enter the image; say the word if you would rather it lived in a separate dev requirements file.

## Cost and timing

Measured locally (Docker Desktop, so its I/O is slower than a runner's):

- Denmark, end to end from the raw file: **57 s** (55 s prefilter, ~1 s exact pass).
- Germany: **524 s** — 503 s prefilter, 21 s exact pass. For scale, the spike's pure-Python filter of the same file was 2,022 s.

Extrapolating to 49 countries (~35 GB of downloads): roughly **2–3 hours of runner minutes per month**, in jobs capped at `max-parallel: 6` so Geofabrik sees a neighbourly six concurrent downloads. The repo is public, so those minutes are free. Peak disk per job is one raw extract (4.8 GB worst case) inside `$RUNNER_TEMP`, deleted the moment it has been filtered.
## Where the artifacts land

A `rail-data-<date>` **pre-release** with the `.pbf` files and `manifest.json` attached — pre-release so it can never displace the app's own latest release. Per the plan: not baked into the image, which would couple a data refresh to a code release and inflate every image.

## Not in this PR, by design

The store, the reader, the builder and anything that consumes a `.pbf` — phase 2, being built in parallel. Nothing here imports from it or anticipates it.

## Test plan

- [x] `python -m pytest tests -q` — 2463 passed, 1 skipped (83 new)
- [x] Every row of the selection mutated and confirmed to fail the suite (table above)
- [x] End-to-end runs of the real script in a container with the osmium CLI, against the actual 494 MB Denmark and 4.5 GB Germany extracts: correct counts, work directory left empty afterwards, two-region `manifest.json` verified
- [x] 20 unresolved relation members sampled and checked against the raw Germany extract with `osmium getid` — none present, so the residual really is cross-border
- [ ] The workflow itself has not run yet — it cannot be dispatched until it is on `main`

## Release notes

- [x] No user-visible change — subject line is enough


rui-nar:
## Adversarial review — not safe to merge as-is

Reviewer ran the real two-step pipeline against eight live Geofabrik country extracts (osmium-tool 1.18), fed the output to the now-merged `src/rail/builder.py` from #348, ran the 83 tests here plus nine independent mutations, and checked `config/rail_regions.yml` against Geofabrik's `index-v1.json`.

### Confirmed correct

- The tag selection matches all four rows of the live Overpass queries exactly — nothing missed, nothing over-wide.
- `is_rail_way` (`scripts/build_rail_extract.py:117`) is character-identical to `src/rail/builder.py:209`.
- The member-way closure is **necessary, not defensive**: `_extract_relation_geometry` returns `None` on a disconnected member graph, so without it those relations surface as "no route found".
- Cross-border residual verified exhaustively on Luxembourg rather than sampled: 10,669 member ways wanted, 1,925 held, 8,744 missing — and **0 of the 8,744 exist in the raw extract**. Nothing is lost to pass ordering.
- Artifacts feed #348's guards cleanly on all eight countries: `missing_nodes=0`, `ways_dropped=0`, `stations_unlocatable=0`.
- The 49-country config is exactly Geofabrik's 53 `europe` children minus the four aggregates.

### Blocking

**B1 — the workflow cannot run.** `.github/workflows/rail-extract.yml:45` installs only PyYAML, but `scripts/build_rail_extract.py:58-60` imports `osmium` and `requests` at module scope. Reproduced in clean `python:3.13-slim`: `ModuleNotFoundError: No module named 'osmium'`, exit 1. `plan` fails, `build` never runs. Every scheduled run produces nothing. Not caught because `test_the_matrix_is_read_from_the_config_not_the_workflow` asserts the command *string* is present, not that it works.

**B2 — a subset rebuild republishes a manifest claiming Europe is one country.** The documented recovery path (dispatch `regions: europe/denmark`) makes `collect_manifest` write a one-region `manifest.json`, and `gh release upload --clobber` overwrites the full one. Same day: 48 `.pbf` assets stay attached but the manifest disowns them. Different day: a new `rail-data-<date>` prerelease holds only Denmark. Related: `publish` runs `if: always()` with no completeness threshold, so 20 transient Geofabrik failures publish a 29-region manifest as current. The fallback for an uncovered region is Overpass — the service that banned the production IP.

**B3 — Phase 1 and Phase 2 disagree about what a valid extract is.** `build_rail_extract.py:333` refuses on `not stations and not ways`; `builder.py:287` refuses on `not counts["ways"]`. The gap is occupied today:

```
europe/liechtenstein  Phase 1: ways=0 relations=10 stations=2 uic_nodes=823 -> PUBLISHES 0.06 MB
                      Phase 2: RailBuildError: no railway ways
europe/andorra  europe/malta  europe/azores  -> RuntimeError: yielded no rail data
```

Cause is legitimate (Liechtenstein's line is `railway=construction` in OSM right now; the other three genuinely have no rail). The defect is that the pipeline has no outcome for "this region has no rail" — so the first scheduled run is 3 red jobs *by design, every month*, making a real failure indistinguishable from expected noise exactly where B2 makes that distinction load-bearing.

### Should fix

- **S1** — the byte-for-byte fixture test is a libosmium version tripwire. Identical within one pyosmium (stable across repeats and Windows/Linux), but 4 bytes differ at offsets 68-75 across versions — the compressed `generator: libosmium/x.y.z` header string. pyosmium 4.0.2 and 4.1.0 DIFFER, 4.3.1 matches; `osmium>=4.0.0` unbounded permits all three. Element counts identical in every case. Compare selection semantically or strip the header.
- **S2** — the fixture contains **no `light_rail` at all**, on either row (way values: rail, narrow_gauge, tram, platform, signal_box, halt; route values: train ×55, railway ×4). Dropping `light_rail` from either constant fails only the parametrized predicate table, which restates the constant three lines above it — so the natural single edit passes all 83 tests. This is precisely the failure mode #349 was about.
- **S3** — two incompatible bboxes ship from one pipeline: `build_rail_extract.py:292` computes it over *every node written*, `builder.py:172` over *rail ways only*. Both are documented as what Phase 3 uses to pick a region. Also: Luxembourg City falls inside 4 configured regions' boxes, Bratislava 3, Zurich 3 — and no tiebreak is defined anywhere.
- **S4** — the download is the one unverified step: no retry, no `Content-Length`, no use of Geofabrik's published `.md5`. Everything downstream is checksummed twice; the input never is. Each transient failure feeds B2.
- **S5** — `requirements.txt` conflicts with `main` (#348 added `osmium>=4.0.0,<5.0.0`; this adds a second unbounded line). Keep the bounded one.
- **S6** — `tests/test_rail_extract_workflow.py` asserts literal YAML strings rather than behaviour. B1 is the proof.
- **S7** — `member_ways_missing` is logged as *distinct way ids* (:346) but the PR body quotes *member slots*; they differ 2.6× on the fixture, and 82% vs 16-19% on Luxembourg. Phase 3 will read the log and get the other number.

### Noted

- "`rail=0` ways stay out of the R-tree" is half wrong — every way is in `way_bbox` deliberately (`relations_near` needs platforms or it loses 12.6% of candidates); they're kept out of *routing* by `_RAIL_IN_BOX`'s `w.rail = 1`. Behaviour holds, stated mechanism doesn't.
- Near-zero regions are worse than zero: Cyprus publishes 2 rail ways under a ~100 km bbox, Iceland 1 way under a 20-metre box. A Cyprus coordinate would select the Cyprus store, find nothing, and degrade to a straight line rather than falling back.
- Peak-disk and cost claims check out (~2-3 h billable job-minutes; free on a public repo).

Issue #366, opened 2026-09-10
Client-side elevation recompute for E2EE activities the #260 backfill must skip


rui-nar:
## Problem

#260 unit 1 fixes elevation gain (`recompute_track_metrics` currently sums every positive delta with no smoothing — a 600 m climb with ±1.2 m of GPS noise reports 4094 m) and backfills existing hand-edited activities with an Alembic data migration.

**That migration cannot correct E2EE users' activities.** `total_elevation_gain` is a plain float column, but the input it must be recomputed from — `elevation_profile_json` — may hold a client-side ciphertext envelope (`is_encrypted_envelope`, `src/utils/encryption_check.py`), and the server cannot decrypt it. The migration detects those rows, skips them and logs the count; they keep the inflated figure indefinitely.

## What is needed

A client-side pass: the Flutter client can decrypt the profile, recompute the gain with the same smoothing parameters as the server, and write the corrected value back.

## The trap

`PUT /api/activities/{activity_id}` (`api/activities.py:1042`) is the obvious place to send it, and it is the wrong one. Its whitelist is deliberately narrow — six E2EE-in-scope columns plus two edit-undo snapshots — and its docstring says so explicitly: *"This is not a general-purpose activity editor."* Widening it to accept `total_elevation_gain` re-opens exactly what that comment is guarding.

Decide the shape first: a dedicated recompute endpoint that takes the derived scalar for an activity the caller can prove it can read, or a separate narrow route alongside the existing one.

## Acceptance criteria

1. An E2EE user's hand-edited activities show corrected climb figures without the server ever seeing plaintext geometry.
2. The smoothing parameters are shared with the server implementation, not duplicated — a divergence here means two users see different numbers for the same track.
3. `PUT /api/activities/{id}` is not widened.
4. Tests cover an enveloped activity end to end.

Split out of #260 (unit 1). Related: #26, #30.

Issue #260, opened 2026-08-25
Import a GPX file directly as an activity (bypass Strava)


rui-nar:
## Problem

Today the only ways to add an activity to a trip are Strava sync and Polarsteps step-import. If a trip is planned in an external tool (e.g. a route drawn in Komoot/RideWithGPS) and the user forgets to record it live with Strava, there is no way to bring that planned track into ViewTrip — the trip is missing an activity with no path to add one manually.

## Feature request

Let a user upload a `.gpx` file and turn it into a normal project activity, alongside the existing Strava and Polarsteps import entry points.

### Acceptance criteria

1. **Menu integration** — "Import GPX file" appears next to "Import activities from Strava" / "Import steps from Polarsteps" in the project toolbar and overflow menu.
2. **Metadata modal** — before/while picking the file, the user specifies: date, start time, end time, and activity type (run/ride/hike/walk/other — same type vocabulary as the rest of the app).
3. **Semantic GPX validation** — the uploaded file is validated server-side before being accepted, not just checked for well-formed XML.
4. **Rejection with a reason** — files that fail semantic validation (e.g. more than one track, no track, too few points) are rejected with a specific, user-readable reason instead of a generic error or silent partial import.
5. **Visual differentiation** — GPX-imported activities are visually distinguishable from Strava-synced ones in the activity list (a small badge on the activity's type icon).

### Out of scope for v1

- Editing/re-importing a GPX-sourced activity's file after creation (use the existing track editor for geometry tweaks).
- Multi-file / batch GPX import.
- Automatic activity-type detection from GPX data.

## Design decisions

- **Provenance field.** `Activity` gains a `source: Optional[str] = None` field (`None` = Strava, as today; `"gpx"` = GPX-imported). Mirrored on `DBActivity` via a new Alembic migration.
- **Synthetic activity ID.** `activity` is a single global table keyed by `id` alone, which is normally the (always-positive) Strava activity ID, and upsert is by `id` with no provenance check. A GPX import allocates a random negative ID (`-secrets.randbits(62)`) with a collision check + retry, so it can never collide with a real Strava ID or another synthetic one.
- **Semantic validation.** Reject: no tracks, more than one track, fewer than 2 points, out-of-range coordinates, a pathological point count. Multiple *segments* within a single track are accepted and concatenated (only multi-*track* files are rejected). Duplicate/non-monotonic timestamps are advisory only (planned routes commonly carry no meaningful per-point time).
[...]

PR #371, opened 2026-09-10
fix(stats): stop counting GPS noise as climbing (#260 unit 1)

rui-nar:
## Summary

Unit 1 of the #260 v2 plan, and the one that is a live bug rather than a missing feature.

`recompute_track_metrics` summed every positive elevation delta, so sensor jitter was booked as ascent. The error only ever adds — each upward flicker counts, no downward one subtracts it — so it accumulates in one direction and grows with the sample count. Measured on a 6000-sample track with a true 600 m climb and ±1.2 m of ordinary GPS/barometric noise:

| method | reported gain |
|---|---|
| before — raw delta sum | **4094 m** |
| 3 m hysteresis alone | 1175 m |
| smoothing alone | 644 m |
| **after** — 11-sample moving average + 3 m hysteresis | **598 m** |

Both halves are needed: noise still crosses a 3 m band often enough over thousands of samples to matter, and smoothing alone leaves residual jitter.

Smoothing is skipped below 20 samples. A window spanning a large fraction of the series does not filter it, it erases it — averaging 100/150/120/170 m yields four identical values and no climb at all, when every step there is real. Below that length the hysteresis band carries the job alone. (This is what the pre-existing 4-point test in `test_track_edit.py` caught, and why it still passes unchanged.)

This only ever applied to gains the app derives itself: hand-edited pieces, splits and GPX imports. Strava-synced activities arrive with their figure already processed — which is why the same climb could disagree several fold between the two sources inside one trip's stats.

## Backfill

A one-off data migration recomputes edited and GPX-sourced rows from their stored profiles.

**Rows whose `elevation_profile_json` holds a client-side E2EE envelope are deliberately left alone.** `total_elevation_gain` is a plain column, but the profile it must be recomputed from cannot be decrypted server-side. The migration detects those rows, counts them and logs the count; the client-side pass is tracked in #366. Not doing this would have zeroed those users' climb figures.

Also untouched: unedited Strava rows (their gain is Strava's), and the `original_*` edit-undo snapshots — those record what the geometry *was*, and are only ever restored through `reset_activity_track`, which recomputes through the fixed path.

## Test plan

- [x] `pytest tests/` — **2650 passed, 2 skipped**
- [x] `alembic upgrade head` from an empty DB, then `alembic check` — no drift; the backfill logs its recomputed / encrypted-skipped / unusable counts. Run against a throwaway DB, not a dev one.
- [x] New `TestElevationGain`: the noisy fixture lands within 25 m of the true 600 m; a companion test pins the old raw sum above 3000 m so a regression cannot pass quietly; short-series, drift-below-threshold, steady-climb, pure-descent and degenerate cases
- [x] New `test_migration_elevation_gain_backfill.py`: edited row corrected, GPX import corrected, unedited Strava row untouched, **E2EE-enveloped row untouched**, idempotent on re-run, malformed profiles skipped rather than zeroed

## Release notes

- [x] `Release-Note:` trailer under the `stats` scope — written for the reader, not the diff: climb totals were counting GPS noise as ascent and existing trips have been corrected.

Part of #260. Follow-up filed as #366.


rui-nar:
## Adversarial review applied

Ran an adversarial review against `96b6dfc` with instructions to falsify rather than confirm. It came back "not mergeable", and it was right on three counts. Fixed in `ae990d4`; I reproduced each finding independently before acting on it.

### The fix had a worse failure mode than the bug, for one class of file

An 11-**sample** window is a different physical width per recording rate: ~60 m of a 1 Hz ride, but a kilometre of a route exported at one point per 100 m. Every hill on a planned-route GPX sat inside the window and was averaged away.

```
sparse planned route, true 600 m  ->    3 m   (as merged in 96b6dfc)
                                  ->  600 m   (now)
```

Planned routes from Komoot/RideWithGPS are precisely what GPX import exists to accept, so this mattered more than the original inflation. The window now spans a fixed 60 m of travel and is skipped when samples are already farther apart than that — a series that sparse is DEM output and carries no sensor noise to remove.

The fixed 3 m band had the mirror problem: on noise-free DEM data anything above ~1 m discards real rollers, while on a bad GPS fix even 3 m leaks phantom ascent. The band is now sized from the noise in the series itself, estimated from second differences (near zero for a smooth series however steep; `sigma*sqrt(6)` for per-sample noise), clamped to [1, 5] m:

| input | band chosen | gain |
|---|---|---|
| clean DEM route | 1.00 m | 600 / 600 |
| 1 Hz ride, sigma 1.2 | 3.54 m | 599 / 600 |
| flat, bad GPS sigma 3.0 | 5.00 m (clamped) | 11 / 0 |

### The migration made a subset of rows worse

`points_to_elevation_profile` stores `0.0` for points with no `<ele>`, so recomputing from such a profile reads each gap as a dive to sea level. Reproduced on a 300-point track at ~500 m with three such points: live import computes **12.1 m**, the migration would have written **152.3 m**. Those rows already hold the honest figure, so the migration now detects the sentinel and skips them. The sentinel itself is filed as #374.

### Corrected activities sat inside uncorrected trip totals

`project.stats_json` caches summed elevation and is recomputed only when NULL. Writing `activity` rows directly left a corrected activity inside a stale total — on the owner's stats screen and on any public share page — until something else happened to touch that trip. The migration now nulls `stats_json` for every project it touched; both readers recompute on NULL. Test added.

### Also

- Distances threaded into the backfill so it filters over the same metres of travel as the live path.
- Documented two deliberate gaps: an edited-then-**reset** activity is indistinguishable from a plain Strava row (`reset_activity_track` recomputes, then clears `is_edited`), so it is skipped rather than risk overwriting Strava's own figures; sentinel rows keep their existing value.
- Prefix sums make the window O(n) instead of O(n * window) — 200k samples at 1 m spacing: 0.16 s, against 0.40 s for the original sample-window version.
- Release note rewritten. The previous one claimed these are "measured the way Strava measures them"; Strava's algorithm is unpublished and this is a box filter plus an adaptive band, so the claim was not supportable.

### Honest limitation, now asserted as a test rather than papered over

Hills only a few metres tall sitting in comparable noise are still under-reported — the band costs roughly its own height per hill. `test_rolling_terrain_is_under_reported_but_not_erased` pins that as a range instead of pretending otherwise. Conservative is the right direction for this figure; what it must never be again is inflated.

### Verification

- `pytest tests/` — **2656 passed, 2 skipped**
- New tests: sparse planned route, clean small rollers, flat ground at sigma 1.2 and 3.0, rolling terrain, no-distances fallback, elevation gaps through `recompute_track_metrics`; migration tests for the sentinel skip and for `stats_json` invalidation
- `alembic upgrade head` from an empty DB then `alembic check` — no drift

rui-nar:
## Correction to two numbers in this PR

A second adversarial pass reviewed the fix commit `ae990d4` itself and cleared it for merge, but caught me publishing figures I had not measured. Both corrected here, verified on this machine against the merged code.

**1. The performance claim was wrong, and backwards.**

`ae990d4`'s commit message says "200k samples at one-metre spacing now cost 0.16 s, against 0.40 s for the original sample-window version". The 0.40 s was a figure from the *first* review that I repeated without re-deriving. Measured directly, same input, 200k samples at 1 m spacing:

| version | time |
|---|---|
| sample-window (`96b6dfc`) | **0.08 s** |
| distance-window, merged (`ae990d4`) | **0.17 s** |

So the change is roughly **2x slower** than what it replaced, not faster. That is the price of a window that means the same thing at every recording rate, and it is worth paying — but the commit message states the opposite and should not be trusted on this point. Prefix sums are still what keeps it O(n): without them the same input took 1.0 s, and the cost would have grown with sample density.

**2. `152.3` is the old algorithm's figure, not the merged one.**

My earlier comment said the migration "would have written 152.3" for a dropout profile. That was measured against `96b6dfc`. Recomputing the same profile with the merged code gives **320.0** against 12.3 m of real ascent — about 26x. The sentinel skip is more load-bearing than I claimed, not less.

Not corrected, because it does reproduce: the `flat, bad GPS σ3.0 → 11 m` row is seed-dependent (the reviewer's own fixture gives 0.0). Both are honest results for their fixture.

Also noted for follow-up: the docstring in `elevation_gain` says the reference fixture "then reports 596 m"; it reports 598.6 m.

Issue #1, opened 2026-05-22
Android web PWA not honoring screen rotate setting


rui-nar:
Installing the web app as PWA in Android is not honoring the setting on rotate/do not rotate screen

Issue #359, opened 2026-09-09
bug: incorrect rail track being create after #345 migration


rui-nar:
Specifically the Paris Gare de l'Est -> Strasbourg is putting the start of the rail track several km south of paris

rui-nar:
## Investigated — reproduced, and it is **not** a #345 regression

Built `europe-france.rail.sqlite` from the `rail-data-2026-09-08` release extract and ran the resolve offline:

```
strategy=relation_endpoints  pts=6567
start [2.3846383, 48.7527843]   ← 13,933 m from Gare de l'Est
end   [7.734263,  48.5851349]   ← 23 m from Strasbourg
km=492
```

Exactly the reported symptom. Then the same resolve through `OverpassRailSource` against live `overpass.openstreetmap.fr`:

```
OVERPASS RESULT strategy=relation_endpoints pts=6567
start [2.3846383, 48.7527843]  start off 13933 m
end   [7.734263,  48.5851349]  end off 23 m
km=492
```

Byte-identical. `git diff 732d17d~1 HEAD -- src/services/overpass_service.py` confirms #345 moved call signatures and changed no strategy logic — `_MAX_SCORE`, `_extract_relation_geometry`, `_nearest_node`, `elements[0]` and `sorted(common)[:10]` are untouched. Phase 3 reproduced a pre-existing bug, which is the parity it was built for.

## Root cause

`_extract_relation_geometry` snaps endpoints with `_nearest_node` over **every** member way of the route relation. Components of relation 5922256 (`TGV InOui 807B : Paris Est → Strasbourg`, `missing_members=0`):

```
graph nodes: 6397
components: 5   sizes: [6292, 37, 27, 22, 19]
start node 48.877186,2.360577 (80 m from the station) → component #1, size 37
end   node 48.585249,7.733966 (6 m from Strasbourg)   → component #0, size 6292
closest gap between the two: 1.30 m
```

That 37-node island is contributed by exactly one member way:

```
way 86199268  {'railway': 'platform', 'area': 'yes', 'name': 'Voies 27 & 28',
               'public_transport': 'platform', ...}
```

A **platform polygon**. It sits on top of the station, so it always wins `_nearest_node` for the origin endpoint; being a closed ring with no connection to track, Dijkstra can never leave it. `_extract_relation_geometry` returns `None`, and every correct Paris Est ↔ Strasbourg relation is discarded the same way:

```
[...]

Issue #207, opened 2026-08-11
Train segment resolve degraded to a straight line with no visible follow-up (Overpass mirrors timing out)


rui-nar:
## What happened

On val (`viewtripweb-val`, `validation-73e9b46`), resolving a train segment (`f0ef7dec-9eb0-47fb-ac66-e57848ffa3a0`) took 176s and came back as a straight line instead of real track:

```
11:59:08  POST .../segments/f0ef7dec.../resolve-route  202
12:00:28  src.services.overpass_service - INFO  - rail strategy B (endpoint relations) failed:
          Overpass query failed (all endpoints): https://overpass.private.coffee/api/interpreter:
          Read timed out. (read timeout=45)
12:02:07  src.services.overpass_service - WARNING - rail geometry resolved: strategy=straight
          points=2 degraded=True elapsed=176.4s
12:02:07  api.segments - INFO - resolve seg=f0ef7dec... type=train strategy=straight points=2
          degraded=True status=resolved
```

A second segment resolved in the same window (a ferry, `2d85e1c8...`) hit the identical
`overpass.private.coffee` timeout on its first strategy at `12:01:53`, but recovered one strategy
later via `way_dijkstra` at `12:03:30` (`degraded=False`) — so this was mirror flakiness, not a
full outage.

## Root cause

`get_rail_geometry` (`src/services/overpass_service.py:76-159`) is working as designed: strategies
A → B → C are tried in order, each retrying across 3 Overpass mirrors
(`_OVERPASS_ENDPOINTS`, `src/services/overpass_service.py:54-58`), and when all of them fail it
degrades to a straight endpoint chord rather than erroring. That's the correct fallback. The bug
is what happens *after* the fallback:

- `pollSegmentResolution` (`flutter_client/lib/src/projects/project_segment_crud_mixin.dart:293-297`)
  gives up after a 2-minute timeout. This resolve took 176.4s server-side — past the poll window —
  so the dialog had already shown "Still resolving route — it will appear shortly" and moved on.
  The user has no further signal from that point.
- The *only* place a degraded result is surfaced today is a one-shot `SnackBar` in
  `flutter_client/lib/src/projects/segment_dialog.dart:515-521`, shown only if the dialog/poll is
  still alive when the job finishes. It wasn't, here.
- `route_degraded` is persisted (`api/segments.py:175`) and threaded into the geo feature
  properties (`flutter_client/lib/src/projects/client_geo_builder.dart:165`), but nothing reads it
[...]

Issue #150, opened 2026-07-28
After resolving a ferry route we should have an editable track where to change or add/remove the track points


rui-nar:
![Screenshot_20260802-202312.png](https://github.com/user-attachments/assets/7230df08-a3c0-4ac4-b55d-c05666e3ac0b)

Sometimes the ferry route won't exist or can be malformed simply because the source data is wrong. We need the ability to manually edit the track to add/remove/move points.

PR #375, opened 2026-09-10
fix(rail): route to the station a relation says it stops at, and re-resolve what an older resolver drew (#363, #364)

### LINKED ISSUE #363: bug: rail legs start short where the station throat is disconnected in OSM (Paris Montparnasse → Bordeaux)
Split out of #359. Same symptom — a rail leg drawn from a point kilometres away from the station — but a different cause, and unlike #359 it is not ours to fix in the graph builder.

## Symptom

Paris Montparnasse → Bordeaux Saint-Jean resolves with its start **3.5 km short** of Montparnasse. After #361 it at least uses the correct relation and the correct line; before, it was a different relation with the same size of error.

```
                        BEFORE #361                             AFTER #361
Paris Mont. -> Bordeaux relation_endpoints start 3415 m 579 km  relation_uic start 3480 m 532 km
```

## Cause: a real gap in OSM's relation membership

All three candidate relations for the UIC pair have `missing_members=0` — the store holds every member way — and **none of them has a platform member**, so #359's fix does not apply. The graph is still in pieces:

```
rel 5928800  TGV 405 : Paris -- Angoulême -- Bordeaux    nodes=21155  comps=4  sizes=[20745, 322, 84, 4]
rel 5928812  TGV 406 : Paris -- Dax -- Hendaye/Tarbes    nodes=22594  comps=3  sizes=[22252, 322, 20]
rel 5928816  OUIGo : Paris − Saint-Pierre-des-Corps …    nodes=20081  comps=2  sizes=[19759, 322]

   start comp #1, size 322 (88 m from the station) ; end comp #0 (22 m from Bordeaux)
   start comp extent  lat 48.8126..48.8405  lon 2.2951..2.3205
```

A 322-node island covering the Montparnasse throat down to the Ceinture, identical across all three relations, simply not joined to the main line by their membership. **Overpass returns the same gap** — this is upstream data, not a store artifact.

Since #361 the resolver snaps both endpoints inside one connected component and picks the main line, which is why the route is now right and only its first 3.5 km is missing. It cannot invent track that no relation lists.

## Why the usual fallbacks do not save it

Strategy C would resolve this correctly from the bbox graph, but never runs: the buffered box is **15.2 deg²** against `_RAIL_BBOX_MAX_AREA = 9.0`, so it straight-lines instead. That cap is a real memory bound (`_MAX_BBOX_VERTICES`), not an Overpass artifact, so raising it is not free.

This is also why `_ENDPOINT_TOLERANCE_KM` is 5 km rather than 2 km in #361 — at 2 km this leg loses its route entirely and degrades to a straight line across France, which is the worse answer.

## Options, roughly in order of honesty

1. **Fix it upstream in OSM.** The relations should list the throat ways. Correct, free, and helps everyone — but slow, and it is one station of many.
2. **Anchor on the relation's stop sequence.** #361 added `relation_node` — every node member in order, with its role and `uic_ref` — precisely for this. The relation *names* Paris Montparnasse as a stop with coordinates; a leg could be routed stop-to-stop and the first hop bridged from the station node to the nearest node of the chosen component. This is the intended consumer of that table and would close the gap without touching the bbox cap.
3. **Bridge components under a small tolerance.** Cheap, and dangerous: the whole reason `_extract_relation_geometry` routes on a graph rather than chaining is that bridging gaps produced the Hanko→Salo teleport (116 km mid-line, 6.2× the real distance). Any bridge needs a hard metre limit and a test that the teleport stays impossible.
4. **Let strategy C run on a bigger box for long legs.** Rejects itself on memory grounds; see above.

Option 2 is the one worth doing, and the data for it is already on the box.

## Reproducing

Build `europe-france.rail.sqlite` from the `rail-data-2026-09-08` extract (recipe in #359 / `tests/fixtures/rail/README.md`), then resolve `{48.8404, 2.3199} → {44.8259, -0.5560}` against a `LocalRailSource`. Report the component sizes of relation 5928800 to see the island.

## Not urgent

The line drawn is the right line on the right route; it starts 3.5 km late. Worth fixing, not worth rushing.

### LINKED ISSUE #364: A resolver fix cannot reach the routes it already got wrong — segments are only re-resolved when they failed, never when they were confidently wrong
Found while shipping #359 / #361.

## The gap

`sweep_degraded_segments` ([src/jobs/route_jobs.py:234](https://github.com/rui-nar/ViewTripWeb/blob/main/src/jobs/route_jobs.py#L234)) retries a resolved segment only when it is flagged `route_degraded` or `route_hafas_failed`:

```python
if seg.route_status != "resolved":
    continue
if not (seg.route_degraded or seg.route_hafas_failed):
    continue
```

Both flags mean *"we resolved around a failure"*. Neither covers *"we resolved successfully and the answer was wrong"*.

#359 was exactly that second case. Paris Gare de l'Est → Strasbourg came back `strategy=relation_endpoints`, `degraded=False`, 6567 points, a plausible 492 km — and started 13.9 km from the station. Nothing about it looked like a failure to any mechanism we have, so:

- the hourly sweep will never pick it up;
- fixing the resolver in #361 changed nothing for the trips already drawn;
- the only remedy is `POST /api/projects/{name}/segments/{seg_id}/resolve-route`, per segment, by hand, for segments nobody can enumerate.

**Every future resolver improvement inherits this.** The better the fix, the more stored geometry it silently fails to reach.

## Why there is nothing to query on

`ConnectingSegment` ([src/models/project.py:123](https://github.com/rui-nar/ViewTripWeb/blob/main/src/models/project.py#L123)) records `route_status`, `route_error`, `route_degraded`, `route_hafas_failed`, `route_degrade_retries`, `route_edited` — every one of them about *whether it failed*, none about *how it was produced*.

`RailGeometry` already carries `strategy` (`relation_uic` / `relation_endpoints` / `coordinate_dijkstra` / `straight`) and it is logged on every resolve, then thrown away. So today there is no way to ask "which segments were resolved by the code that had the platform bug", or even "which segments used strategy B".

## Proposal

**Stamp provenance on the segment.** At minimum a resolver version — an integer bumped by hand when a change alters geometry quality — and the winning `strategy`, which costs nothing since it is already computed:

```python
route_resolver_version: int = 0
route_strategy: Optional[str] = None
```

**Extend the sweep to stale stamps.** A segment whose `route_resolver_version` is below the current one is re-resolved, at a deliberately low rate, alongside the degraded ones. That makes a resolver fix reach existing data on its own, which is the whole point.

Three constraints it must respect:

- **`route_edited` is untouchable.** A hand-drawn track (#150) must never be silently re-resolved; the existing guard has to cover this path too.
- **Rate and budget.** With `RAIL_SOURCE=local` a re-resolve is cheap, but on Overpass it is not, and a version bump would otherwise queue every rail segment in the deployment at once. `MAX_DEGRADE_RETRIES` and the existing hourly cadence are the model; a stale-stamp sweep probably wants a smaller per-run cap than the degraded one.
- **Only bump the version for changes that actually alter geometry.** A stamp that moves on every release is a permanent re-resolve treadmill.

**A one-off admin path** would also have helped here and is much smaller: re-resolve every rail segment in a project, or every segment matching a strategy, on request.

## Smaller alternative, if the above is too much

Persist `route_strategy` alone and nothing else. It does not automate anything, but it turns "which trips are affected by this bug" from unanswerable into one query — which, on the day #359 was found, is what was actually missing.

rui-nar:
Closes #363. Closes #364.

Both fell out of #359 and are the two halves of the same lesson: a rail leg can be drawn wrongly without anything failing, and once drawn, nothing goes back for it.

## #363 — legs starting short where OSM leaves the station throat disconnected

Paris Montparnasse → Bordeaux was drawn from the Petite Ceinture, 3.5 km out. None of #359's fixes reach it: all three candidate relations hold every member way and carry no platform member. Their graphs are genuinely in pieces — a 322-node island at the Montparnasse throat — and Overpass returns the same gap, 100 m wide at the point where the throat meets the Ceinture.

The resolver now uses **the relation's own stop sequence**. Where a relation names a stop at a leg's endpoint, the component holding that stop may be joined to whatever it comes within `_COMPONENT_BRIDGE_M` of — and nothing else may.

The anchor is the load-bearing part, not the bridge. A blanket 250 m join would let a relation that does *not* serve this leg stitch itself into reach of `_ENDPOINT_TOLERANCE_KM` — the guard #359 had to widen to 5 km for exactly this leg. Anchoring means a bridge is only ever built toward a station the relation itself asserts it calls at, so a relation that clears the endpoint check after bridging is one that actually serves the endpoint.

`_COMPONENT_BRIDGE_M = 250.0` is measured, not fitted. Across 400+ France relations, 162 have graphs in pieces; their 868 non-main components sit ≤10 m (505), ≤100 m (43), ≤250 m (6), >1 km (50) from the rest. There is no cliff to read a limit off, so 250 m is the p95: it clears the reported 100 m with headroom, excludes the kilometre tail, and is three orders of magnitude below the 116 km Hanko→Salo teleport that graph routing exists to prevent.

**`RailStore.relation_geometry` now emits node members** in the exact shape Overpass's `out geom` uses — verified against the live API first, not assumed: relation 5928800 returns all 16 node members as `{"type":"node","ref","role","lat","lon"}`. So `_extract_relation_geometry` reads `rel["members"]` with no branch, no `RailSource` ABC change, and both sources are fixed together. This is the first consumer of the `relation_node` table #359 added, so **no schema bump and no region rebuild**.

## #364 — a resolver fix could not reach the routes it already got wrong

Every flag a segment carried recorded whether a resolve *failed*. #359 came back `strategy=relation_endpoints`, `degraded=False`, a plausible 492 km — and started 13.9 km from the station. `sweep_degraded_segments` will never look at that, so #361 fixed the resolver and changed nothing for trips already drawn.

Segments now record how their geometry was *produced*: `route_resolver_version` (a hand-bumped resolver generation) and `route_strategy` (the winning strategy, previously computed, logged and discarded). A new hourly `sweep_stale_resolver_segments` re-resolves rail segments stamped below `RESOLVER_VERSION`.

- **Capped at 3 per run**, in Overpass slot-seconds rather than wall clock: concurrency is 1, one rail resolve issues 4–5 queries, and the measured endpoint numbers put a healthy query at 8.4 s and an unreachable host at the full 60 s timeout. So 3/hour is ~3% of a good hour and ~25% of a bad one, drains a few-hundred-segment deployment over a couple of days, and slow is the point — a version bumped by mistake is noticed long before it has re-resolved much. This IPv4 was hard-banned by overpass-api.de in September for exactly this shape of traffic.
- **Cron on a fixed minute, not `interval`**, so it does not sit permanently in step with the degraded sweep competing for the single Overpass slot.
- **`route_edited` is guarded explicitly.** That guard is *not* inherited — the track-edit endpoint clears both provisional flags, so a hand-drawn track is invisible to the degraded sweep by accident of the flags rather than by a check. A stale stamp has no such accident, and without this line the first version bump would discard every hand-drawn track in the deployment (#150).
- **No stamp means version 0**, deliberately, not "unknown, leave alone" — the routes #359 got wrong are exactly the unstamped ones, and the other reading would help only future bugs.
- **No migration**, verified rather than assumed: `segment_json` is free-form TEXT and the fields round-trip through `from_dict`.

`RESOLVER_VERSION` carries the rule for bumping it, because a stamp that moves every release is a permanent re-resolve treadmill against a rate-limited API.

### Pre-existing data-loss bug fixed along the way

`ProjectRepo`'s row mapper omitted `route_degrade_retries`. That is destructive, not merely lossy: `save_project` rewrites every item's payload from `to_dict()`, so any structural write — reordering a trip, adding an item — reset the #207 retry budget to 0. I verified this independently on `main` before accepting the fix. The provenance stamp would have shared the same fate.

## Measured

Seven French legs on the real France store built from `rail-data-2026-09-08`, resolved through `LocalRailSource` with `_overpass` monkeypatched to raise, so nothing reached the network. Re-run by me on the **merged** branch, which is the combination neither agent could test:

```
                              before            after
Paris Est -> Strasbourg       94 m,  438 km     94 m,  438 km    (#359, unchanged)
Paris Mont. -> Bordeaux     3480 m,  532 km    119 m,  536 km    <- #363
Paris Lyon -> Lyon Part-Dieu   30 m,  429 km     30 m,  429 km
Paris Nord -> Lille Flandres   38 m,  250 km     38 m,  250 km
Strasbourg -> Colmar           48 m,   66 km     48 m,   66 km
Lyon PD -> Marseille SC        55 m,  325 km     55 m,  325 km
Nantes -> Rennes               89 m,  350 km     89 m,  149 km   <- see below
```

**Nantes → Rennes moved and it is a fix, not a casualty.** `TER 4 : Rennes - Nantes` is disconnected at its own Nantes throat, so its best piece started 19,309 m out and was refused; the winner was `TGV 733 : Rennes/Nantes -- Le Mans -- Lyon -- Marseille`, a split train carrying both branches, so routing Nantes→Rennes inside it detoured via Le Mans — 350 km against a 99 km crow. Bridged, `TER 4` routes from 37 m and gives the 149 km line through Redon.

No polyline gained a jump: longest segment unchanged on six legs, 995 → 1009 m on Montparnasse (both mapped LGV straights). Resolve cost +0.9%.

## Tests

**Full suite on the merged branch: 2682 passed, 3 skipped, `PYTEST_EXIT=0`.**

- #363: 29 new tests, each verified to fail by actually reverting the change — reverting the resolver bridge fails 12, reverting the store's node members fails 3. Includes a test that names a stop at each end of the Hanko→Salo shape and proves 780 km still cannot be bridged, and the `test_rail_source.py` parity oracle now emits node members so it still stands in for Overpass.
- #364: 12 new tests covering the unstamped default, the `route_edited` guard, version comparison, the per-run cap actually capping, the stamp write, the row mapper, and the manual-edit strategy.
- New fixture documented in `tests/fixtures/rail/README.md`, including the one recipe change from #359's: junction vertices must be kept or the graph shatters into 25 components.

## Known limitations, deliberately not fixed here

- `RESOLVER_VERSION` is stamped on every resolved segment including ferry and bus, but the sweep filters to `segment_type == "train"`. A future ferry-resolver fix would mark ferry segments stale and never sweep them. Harmless today.
- `_best_relation_geometry` ranks candidates purely on endpoint proximity, tying three relations at 0.000000 on Nantes→Rennes. Pre-existing; the winner is deterministic by relation id, but it is a real fragility.
- A stale re-resolve run while every Overpass mirror is down will downgrade good track to a straight chord. It is stamped anyway so the sweep terminates in one attempt per segment per bump, and `sweep_degraded_segments` then owns it with its own budget to restore real track.


rui-nar:
## Adversarial review round — three defects found and fixed (`067087d`)

An independent hostile review found three real problems, two of them in claims the code made about itself. All fixed; nothing was waved through.

### 1. The stale sweep would have downgraded most train segments in the deployment — CONFIRMED, severe

`_compute_segment_geometry` asks MOTIS for a matched train and returns the trip's own track as `motis_trip`, **never reaching the rail resolver**. So `RESOLVER_VERSION` said nothing about that geometry, and re-drawing it was not merely wasted work:

- MOTIS answers from a departure board, which holds nothing for a train that ran months ago → `HafasError`;
- the verdict persisted `route_hafas_failed` with a "Train lookup failed" message the client renders, and **replaced the trip's real track with a line between the two endpoints**;
- the segment was then provisional, so the *uncapped* `sweep_degraded_segments` inherited it for `MAX_DEGRADE_RETRIES` more attempts that failed identically — six resolves per historical segment.

This app records past trips, so that is most train segments in a deployment. A fix meant to reduce traffic would have produced precisely the shape that got this IP banned. The PR's budget arithmetic counted only the stale sweep and only Overpass.

**Fixed with two independent guards, either sufficient.** `route_resolver_version` is now stamped only when a strategy in `RAIL_RESOLVER_STRATEGIES` produced the geometry, and the sweep selects only segments that strategy says the rail resolver drew. For rows written before the stamp the strategy is unknown, and unknown resolves to a *refusal* in the one direction where a wrong guess destroys data — a `train_number` means MOTIS was tried. The routes #359 drew wrongly have no train number, so the population this sweep exists for is untouched.

### 2. The named-stop rule did not restrict what it claimed — CONFIRMED

A home component bridges to every component it approaches, and the far endpoint's stop normally sits on the main line — that end is usually the one that is *not* broken. So the main line became a home and was joined to every stray fragment within 250 m along its entire length: the general gap-chaining the rule exists to prevent, reached through the door marked "named stop".

The review demonstrated it by deleting the Montparnasse stop node and getting the identical bridge. **The largest component is never a home now** — a stop already sitting on the route has no gap to close.

### 3. An invented path is bounded at two edges, not one — CONFIRMED, minor

Two broken throats either side of a shared fragment each bridge to it. Still ≤250 m per edge, so still three orders of magnitude from the Hanko→Salo teleport — but the docstring said "cannot accumulate into a path", and that was wrong. Bound restated honestly.

### Verified sound

`_requeue_resolve` is a faithful extraction of the degraded sweep — same CAS keys, same ordering, same exception scopes, #207's budget intact. The two sweeps' candidate sets are genuinely disjoint. The row-mapper data-loss fix is real.

### Re-measured after the fixes

Seven French legs on the real France store, network blocked — **byte-identical to before the review**, so tightening the anchor cost none of the fix:

```
Paris Mont. -> Bordeaux    119 m, 536 km      Nantes -> Rennes   89 m, 149 km
Paris Est -> Strasbourg     94 m, 438 km      four others        unchanged
```

**Full suite: 2690 passed, 3 skipped, `PYTEST_EXIT=0`** (8 new tests). Each new guard verified to fail with the guard reverted — including the two anchor tests, which fail on the pre-fix code while the other 19 in that file pass either way.

### Carried forward, not fixed here

- `sweep_orphaned_jobs` runs only at API startup, so a segment left `pending` by an enqueue failure after a successful compare-and-set waits for a restart. Pre-existing for the degraded sweep; this PR widened the exposure, and fix #1 above narrows it back to segments the rail resolver drew. Worth its own issue.
- Ferry/bus stamp the segment type as their strategy rather than the inner strategy actually used, so the "which strategy drew this" query is uninformative for them. No wrong answer today.

rui-nar:
## Second review round, on the fix commit itself — three more found, all fixed (`4f778d2`)

Re-reviewing `067087d` rather than stopping at one pass was the right call: two of the three findings were defects *in the fix*.

### 1. The MOTIS guard was keyed on the wrong variable — CONFIRMED, severe

The hazard is not "the rail resolver did not draw this", it is **"re-resolving asks MOTIS about a train that ran months ago"**. That is decided by the `train_number` the sweep forwards, not by `route_strategy`.

And a rail-drawn segment can carry one: MOTIS matches the train, returns a trip with no shape (`_trim_polyline` returns `[]` below two points), and its *stops* go to the rail resolver — rail strategy, `hafas_failed=False`, `train_number` set. My guard admitted exactly that population, and a test asserted it as correct. Re-resolving one replaces a track drawn through MOTIS's full stop list with a two-endpoint chord, flags it failed to the user, and hands it to the uncapped degraded sweep. The same six-resolve shape the guard existed to stop, deferred to the next `RESOLVER_VERSION` bump. The reviewer reproduced it end to end.

Two real paths into that population: the no-shape case above, and `update_segment` adding a train number to an already-resolved rail segment without touching its polyline or stamp.

**The guard now refuses on the train number.** The cost is stated in the code rather than hidden: train segments drawn by the rail resolver between its arrival and the MOTIS switch carry a train number, and this sweep will never reach them. They are the user's to re-resolve by hand — the trade against silently replacing their track with a chord.

### 2. The write-side stamp gate had no test — CONFIRMED

My previous commit message said every guard was verified by reverting it. That was false for this one: widening the strategy set in `api/segments.py` left all 67 tests green, because the test that claimed to cover it inspected the constant instead of driving a resolve. It now runs a real `_resolve_route_job` and reads back what was stored. Reverting the gate fails it.

### 3. The main-line rule is reverted — CONFIRMED regression

`max(comps, key=len)` approximates "the component the route runs on" by node count, and a station throat is denser than a short line. Measured against the France store: **61 relations changed**, and two `route=train` relations lost a correct path — 9755791 and 9755793, `TER 02` out of Valence, ending **48.67 km** and **60.49 km** short, rejected by `_ENDPOINT_TOLERANCE_KM` and masked end-to-end only by another relation serving the same UIC pair.

That traded round one's unproven risk for a confirmed regression. Round one offered two remedies and the other one is right: **say plainly what the rule guarantees** — at most two invented edges, each at most `_COMPONENT_BRIDGE_M` — and stop claiming it only bridges toward a named stop. The grade-separation risk is now documented as known and unmitigated; distinguishing it from a genuinely missing junction needs `layer`/`bridge`/`tunnel` tags the graph does not carry.

Both relations are back to 179.2 km and 233.9 km with exact endpoints.

### Also fixed

`test_an_invented_path_is_bounded_at_two_edges_not_one` asserted `<= 2` on a graph that built one edge — and passed with bridging deleted entirely. It now exhibits the two-edge path and measures both spans.

### Verified sound this round

Termination: no re-selection loop is constructible — every verdict for a swept segment either stamps, writes a non-rail strategy the guard rejects, or fails. `main` did not change identity between endpoint iterations. Params passthrough is still covered.

### Re-measured

```
Paris Mont. -> Bordeaux    119 m, 536 km        Nantes -> Rennes   89 m, 149 km
Paris Est -> Strasbourg     94 m, 438 km        four others        unchanged
TER 02 Valence-Annecy     179.2 km, ends exact  Valence-Geneve     233.9 km, ends exact
```

**Full suite: 2689 passed, 3 skipped, `PYTEST_EXIT=0`.** Both new guards verified failing when reverted, individually.

rui-nar:
## Both carried-forward items closed (`c9738bc`) — no loose threads left

### 1. Segments stranded "pending" with nothing coming for them

`sweep_orphaned_jobs` asks whether a **job** is owed, and only at startup — deliberately, since at startup nothing is in flight and it need not tell a live job from a dead one. Two holes follow:

- `_resolve_route_job` marks its job **done** and returns *without writing a verdict* when the project cannot be loaded (a rename mid-resolve). The segment then waits on a terminal job row that no sweep looks at again — recovered by nothing, ever.
- An enqueue that fails after the compare-and-set is recovered only by the next API restart.

Until now the only thing that noticed either was the Flutter client's five-minute check, and only if someone reopened the project.

`sweep_stuck_pending_segments` asks the opposite question — *is this segment waiting for something that is never coming* — so it closes both. It is safe against a serving API, which the startup sweep is not, because it skips anything holding a non-terminal job row: a genuinely running resolve has one (`mark_running`), so live work is never duplicated, and a merely lost job stays the startup sweep's to retry with its own `MAX_ATTEMPTS` budget this must not spend.

`STUCK_PENDING_AFTER_S` is 15 minutes against a ~5 minute worst case — four or five Overpass queries at `_TIMEOUT_HTTP` across two endpoints — plus queue wait. An unparseable or missing timestamp counts as **not** stuck: a segment we cannot date is one we cannot prove is stuck, and re-queueing on a misparse would fight a resolve that is still alive. Hourly at :40, on its own minute for the same reason as the other two sweeps.

### 2. Ferry and bus now record the strategy they actually used

`_get_route_geometry` picked between `relation`, `way_dijkstra` and `ferry_yes_dijkstra`, logged the choice, and dropped it — the segment was stamped with the *mode* instead (`"ferry"`). So the one question `route_strategy` exists to answer was unanswerable for anything that isn't a train, which was half the point of adding the field.

Now returned and stored, qualified by mode (`ferry_relation`, `bus_way_dijkstra`) because "relation" means something different for each. Nothing keys on the old values — they are provenance — and rows already written keep `"ferry"`/`"bus"`.

### Verification

Seven new tests, each verified against the guard it covers rather than assumed:

- dropping the non-terminal-job check → `test_a_live_resolve_is_never_duplicated`, `test_a_lost_job_is_left_to_the_startup_sweep` fail;
- dropping the age check → `test_a_recently_started_resolve_is_left_alone`, `test_an_undateable_timestamp_is_not_assumed_stuck` fail.

**Full suite: 2696 passed, 3 skipped, `PYTEST_EXIT=0`.** Rail geometry unchanged on real France data — Paris Est → Strasbourg 94 m / 438 km, Montparnasse → Bordeaux 119 m / 536 km, Nantes → Rennes 149 km.

Every finding from both review rounds is now either fixed or, for the grade-separation case, documented in the code as known and unmitigated with the reason it cannot be told from a genuinely missing junction.

Issue #374, opened 2026-09-10
Elevation profiles store 0.0 for points with no <ele>, inflating any later recompute ~12x


rui-nar:
## Problem

`points_to_elevation_profile` (`src/models/track_edit.py:117`) stores `0.0` for any point that carried no elevation:

```python
elevations = [p.elev if p.elev is not None else 0.0 for p in points]
```

A track through the Alps with three `<ele>`-less points is therefore stored as diving to sea level and climbing back out. Nothing in `validate_for_import` rejects a partially-elevated GPX, so these rows exist.

Found while reviewing #260 unit 1 (PR #371).

## Two consequences

1. **Anything recomputing from the stored profile inflates.** Measured on a 300-point track at ~500 m with three missing points: the live import path reports **12.1 m** of gain (it drops elevation-less points), while recomputing from the stored profile gives **152.3 m** — 12x. PR #371's backfill migration detects the sentinel and skips those rows rather than corrupt them, so this is contained for now, but the trap stays armed for the next thing that recomputes.
2. **The elevation chart draws the dive.** The profile is what the client plots, so those activities render a spike to zero.

Editing such an activity today already produces the inflated figure, via `align_points` interpolating over the `0.0` values — so this is pre-existing, not introduced by #371.

## Options

- Interpolate elevation across gaps when building the profile (matching what `align_points` does when reading it), so the stored series has no holes and no sentinel.
- Or carry `null` through the profile and teach every reader to skip it — more honest, more places to touch.

Either way a repair pass is needed for rows already stored with the sentinel, and it should be paired with lifting PR #371's skip so those activities finally get corrected gains.

## Acceptance criteria

1. A GPX with some points lacking `<ele>` stores a profile with no fabricated sea-level samples.
2. Recomputing gain from a stored profile matches what the live import computed for the same track.
3. Existing rows carrying the sentinel are repaired, and #371's `_has_elevation_dropout` skip is removed in the same change.
4. The elevation chart shows no dive for those activities.

rui-nar:
Fixed in #382 (merged as d83d73d).

`points_to_elevation_profile` now interpolates across points with no `<ele>` instead of storing `0.0`, and migration `c4a9e1f70b38` repairs already-stored rows — filling the holes, recomputing the gain, rebuilding the low-res copy the chart loads first, and nulling cached trip totals. Recompute-from-storage on the reference track went from **232.6 m** to **14.2 m**, against the live import path's 14.4 m.

[...]

PR #382, opened 2026-09-11
fix(stats): measure elevation noise, and stop storing sea level for a missing reading (#376, #374)

### LINKED ISSUE #376: Elevation gain: noise inferred from sampling geometry misjudges sparse-noisy, held-altitude and curved-route tracks
## Problem

PR #371 replaced raw positive-delta summation with a distance-spanned moving average plus a noise-adaptive hysteresis band. It is better than what it replaced on every input tested, and it shipped — but a second adversarial review showed **two of its stated premises are false**, with measurable residual error on real-world inputs.

### Premise 1: "sparse spacing implies route-planner DEM output, which has no sensor noise"

`_smooth_elevations` (`src/models/track_edit.py:172`) skips smoothing entirely when median sample spacing >= 30 m, leaving the band (capped at 5 m) as the only defence. 5 m is under 2 sigma for GPS-only altitude. Measured, 30 km of flat ground at 40 m spacing:

| noise | reported climb (true 0) |
|---|---|
| sigma 2 | 250 m |
| sigma 3 | 652 m |
| sigma 5 | 1720 m |

Real sources: a non-barometric phone or Garmin using smart recording, or a track simplified before upload. Spacing does not identify the source, which is the flaw.

### Premise 2: "second differences are near zero for a smooth series, however steep"

True for straight flanks, false for curvature — second differences scale with spacing squared. A *clean* curved route at >=30 m spacing therefore hits the 5 m clamp, and rollers up to ~2x the band are erased because the reference sits mid-oscillation:

| clean sinusoidal rollers | true | reported |
|---|---|---|
| 10 m over 600 m, 100 m spacing | 500 m | **0** |
| 6 m over 400 m | 450 m | **0** |
| 4 m over 500 m | 240 m | **0** |

`test_rolling_terrain_is_under_reported_but_not_erased` passes only because its fixture uses triangles, which have zero curvature on their flanks. Make the hills sinusoidal and it fails at 100 m spacing.

### Premise 3 (same root): held/repeated altitude values collapse the estimate

Devices that update altitude less often than position emit runs of identical values, so the median second difference goes to 0 and the band drops to its 1 m floor while the steps between updates stay large. 30 km flat, altitude held for 5 samples: sigma 1.2 -> 155 m, sigma 2 -> 347 m, sigma 3 -> 634 m. Plain quantisation (0.1/0.2/1 m) does **not** trigger this — results stay within 2% of unquantised.

## Direction

The common cause is inferring noise from sampling geometry. Candidates:

- Estimate noise from a robust high-frequency statistic that survives runs and curvature (e.g. MAD of residuals against a local *linear* fit rather than second differences, or a run-length-aware estimator).
- Decouple the smoothing decision from spacing: smooth by distance whenever there are enough samples in the span, regardless of median gap, and let the noise estimate decide the band.
- Consider carrying a per-activity provenance hint — a GPX with no `<time>` and regular spacing really is a planned route, and we will know that once #260 unit 3 lands, instead of guessing from geometry.

## Acceptance criteria

1. 30 km of flat ground at 40 m spacing with sigma 3 noise reports < 50 m, not 652 m.
2. Clean sinusoidal rollers (10 m over 600 m at 100 m spacing) report most of their 500 m, not 0.
3. A held-altitude series (value repeated 5-10x) reports < 50 m on flat ground.
4. The rolling-terrain test uses sinusoidal hills, not triangles.
5. No regression on the cases #371 already fixed: 1 Hz noisy climb ~600, sparse clean route ~600, flat sigma 1.2 ~0.
6. While here: `elevation_gain`'s docstring says the reference fixture "then reports 596 m"; it reports 598.6 m.

Follow-up to #371 (issue #260 unit 1). Not a regression — every case above is at least as good as the raw-sum code this replaced; the sparse+noisy+flat class is worse than the intermediate `96b6dfc` only, which was itself destroying 70-90% of real climb on those tracks.

rui-nar:
## Summary

Closes #376 and #374 — the two follow-ups that came out of the adversarial reviews of #371, cleared before starting #260 unit 2.

Both land together because they are the same bug seen twice: the app inferred things about elevation data from its shape, and guessed wrong.

### #376 — stop guessing noise from sample spacing

`elevation_gain` decided how much to filter from the gap between samples, and sized its hysteresis band from median second differences. Both premises were false:

- **"Sparse spacing means route-planner DEM output, so no noise."** Spacing does not identify the source — a non-barometric phone, a Garmin on smart recording, or a track simplified before upload is sparse *and* noisy, and skipping the filter let the raw noise through.
- **"Second differences are near zero for a smooth series."** True only for straight flanks. Second differences scale with the square of the spacing on a bend, so a *clean* curved route measured as noisy and had its rollers erased. Devices that hold an altitude reading across several GPS fixes broke it the other way: runs of identical values drive the median difference to zero, the band collapses to its floor, and every step between updates counts as climbing.

One measurement now replaces both guesses. Noise is read from median absolute k-th differences over orders 3-5, at a stride equal to the median run of repeated values, keeping the smallest estimate — differencing k times annihilates any local polynomial below degree k, so terrain must bend harder than that to register, while noise always survives at `sigma*sqrt(C(2k,k))`. The run stride is what survives held readings. Smoothing treats 60 m as a floor and widens to hold `(sigma/1.0)^2` readings, capped at 240 m; the band is sized from *residual* (post-smoothing) noise, clamped [1, 20].

| case | before | after |
|---|---|---|
| 30 km flat, 40 m spacing, sigma 3 (true 0) | 675 m | **0 m** |
| clean sinusoidal rollers, 10 m/600 m @ 100 m (true 500) | 253 m | **500 m** |
| held-altitude flat, sigma 3 (true 0) | 680 m | **0 m** |
| 1 Hz noisy climb (true 600) | 598.6 m | 599.2 m |
| sparse clean planned route (true 600) | 600 m | 600 m |
| 200k samples | ~0.19 s | 0.189 s |

Every constant came from a 20-seed sweep; the sweep figures are recorded in the docstrings rather than in a commit nobody will find again.

**The rolling-terrain test used triangular hills** — straight flanks, zero curvature, so it could never have caught this. It is sinusoidal now.

### #374 — stop storing sea level for a missing reading

`points_to_elevation_profile` wrote `0.0` for any point with no `<ele>`, so a track through the Alps with three such points was stored as diving to sea level and climbing back out. The chart drew the dive, and anything recomputing gain from storage inflated: measured on a 300-point track at ~500 m, the live import path gives **14.4 m** and a recompute from its own stored profile gave **232.6 m**.

Gaps are now interpolated by cumulative distance — matching what `align_points` already derives when reading a profile back, keeping the array numeric for the chart and the low-res downsample. Recompute-from-storage now gives **14.2 m** against the live path's 14.4 m.

A new migration `c4a9e1f70b38` repairs stored rows: it fills the holes, recomputes the gain, rebuilds the low-res copy (the chart loads that first — leaving it would keep drawing the dive), nulls cached `stats_json` for affected trips, skips E2EE envelopes, and is idempotent. These are exactly the rows `b7f1a3c9d204` detected and deliberately skipped, so they finally get corrected figures.

`b7f1a3c9d204` is merged and may be applied in production, so it was **not** edited — #374's acceptance criterion 3 asked for its skip to be "removed in the same change"; the new migration picks up the rows it skipped instead. Same effect on data, different mechanism.

### One tightening during integration

The repair initially treated *every* exact `0.0` as a sentinel whenever the series' median sat above 50 m — which is the test `b7f1a3c9d204` used to decide a row was unsafe to **recompute**. Repairing is destructive where skipping was not: a ride down from the Alps to a Mediterranean beach has both a high median and genuine `0.0` samples, and blanket-filling would have interpolated a real coastline away, with no downgrade to undo it.

Detection is now per run of zeros, judged by that run's own neighbours: a sentinel leaves a cliff (the samples either side sit hundreds of metres up), while real sea level is walked down to. Strictly narrower than the old test, which is the safe direction — every row this repairs is one `b7f1a3c9d204` skipped, and a row it declines to touch keeps the honest figure the live path gave it. Test added, and confirmed failing against the pre-tightening migration.

### Known limits, stated rather than hidden

- A track that is sparse **and** noisy **and** genuinely hilly is under-reported: 20 m hills at 100 m spacing under sigma 2 read 349 m of a true 600. You cannot both filter sigma-2 noise and preserve 20 m features at that sampling. It beats the 236 m a raw-sigma band gave, and beats an inflated answer.
- Rows whose `distances_km` and `elevations_m` lengths disagree are skipped, not repaired — interpolation cannot proceed without trustworthy distances. Covered by a test.

## Test plan

- [x] `pytest tests/` — **2731 passed, 2 skipped**
- [x] `alembic upgrade head` from an empty throwaway DB, then `alembic check` — "No new upgrade operations detected". Never run against a dev DB.
- [x] Every new #376 test confirmed **failing against `main`** by stashing the source change; the two that do not fail (quantisation, linear-time) guard the new logic rather than the old bug, and are labelled as such.
- [x] The integration tightening's test confirmed failing against the pre-tightening migration.

## Release notes

- [x] `Release-Note:` trailers on both fix commits, under the `stats` and `gpx` scopes.


rui-nar:
## Adversarial review applied (`e3eca5a`)

Fable review of this PR. Verdict was "not a clean merge as-is, but close" — four findings acted on, one filed, one disputed. I reproduced each before acting.

### The sentinel test was wrong in both directions — altitude is the wrong axis

**Too narrow.** A Dutch or Danish track at 30 m has a median below any sea-level margin, so its dropouts were never detected. That mattered more than it looks: `b7f1a3c9d204` ships in the same release and runs *first*, its own median test also passes, so it recomputed those rows **from the fabricated dive**. Measured on a 3 km track at 30 m with three `<ele>`-less points:

```
live import path            1.01 m
b7f1a3c9d204 would write   18.66 m
this pass (before)          declined to repair
```

Neither migration has a downgrade, so that would have shipped as permanent.

**Too wide.** A route along 2 km of DEM beach at exactly 0.0 before climbing a 60 m coastal cliff has one neighbour far above the margin, so the whole beach was flagged and interpolated into a ramp. DEM sea level *is* exactly 0.0 — that is the normal shape of a coastal planned route, not a corner case.

Detection is now by **grade**: the sentinel replaces a reading and leaves a step no terrain can hold (30 m over 10 m of travel is 300%), while real ground is arrived at, whether that is a beach at 0 m or a canal path at 30 m. Steps under 5 m are not consulted, so two samples a centimetre apart cannot look like a cliff. Seven cases checked, both new ones tested.

### The constants cited a fixture that does not reproduce

The review could not re-derive "349 m of the real 600" for the window cap. Neither could I — the fixture as described gives **239 m**. Both docstrings now spell the fixture out precisely enough to re-run, with measured means and ranges over 20 seeds:

| | measured |
|---|---|
| 240 m window cap (30 sine hills, 20 m/600 m, 100 m spacing, sigma 2, true 600) | 239 m (218-254) |
| same, uncapped widening | 108 m (4-252) |
| 4.5x band (30 km flat, 40 m spacing, sigma 3, true 0) | 3 m (worst 28) |
| same at 3x / 4x / 5x | 25 / 8 / 0 m |

### The timing test measured the machine, not the code

`test_long_track_stays_linear` asserted a wall-clock bound — and duly failed on my box while the full suite was running (21 minutes instead of 5). It now compares 50k against 200k samples and asserts the *ratio*: linear is ~4x, quadratic would be ~16x.

### Filed, not fixed: #386

The band is sized from a **white-noise** model, and k-th differences only see the innovation of an autocorrelated series. GPS altitude noise is strongly autocorrelated, so on a phone-without-barometer recording at 1 Hz this PR leaks *more* phantom climb than the code it replaces — AR(1) flat ground, 30 km, marginal sigma 4: **807 m vs main's 468 m**. Barometric-class noise is unaffected.

This is the one case where the PR is worse than `main`, and both remain far better than the raw-sum code actually in production. Fixing it properly means detecting correlation rather than assuming independence, which is a redesign of the estimator, not a constant tweak — so it is #386 with the measurements, not a patch bolted onto this branch. #386 also asks for the tuning fixtures to be committed and runnable, which is how the unreproducible figures above happened in the first place.

### Not acted on

The review's F5 (a JSON `null` inside a profile would abort `alembic upgrade head`) is real but **pre-existing and shared** — `b7f1a3c9d204`'s `sorted()` raises first, on `main` today — and no writer in the codebase can produce such a row. Guarding it belongs with the other migration-robustness work, not here.

### Verification

- `pytest tests/` — **2733 passed, 2 skipped**
- `alembic upgrade head` from an empty DB then `alembic check` — no drift
- The grade fix's two new tests confirmed failing against the previous commit

PR #389, opened 2026-09-11
fix(stats): measure correlated GPS noise, not just its innovation (#386)

### LINKED ISSUE #386: Elevation band assumes white noise; correlated GPS altitude noise still reads as climbing
## Problem

`elevation_gain`'s hysteresis band is sized from a **white-noise** model: `_noise_threshold` derives the residual left after smoothing as `sigma / sqrt(window / stride)`, where `sigma` comes from median absolute k-th differences.

k-th differences only see the *innovation* of an autocorrelated series. GPS altitude noise is strongly autocorrelated — satellite geometry and multipath drift over minutes, not samples — so for that input `sigma` is measured far below the series' actual marginal deviation, the window stays at its 60 m floor, the residual model predicts ~0.15 m, and the band collapses to its 1 m floor. Meanwhile the smoothed series really is wandering by metres, and every wander is counted as climbing.

Found by adversarial review of #382. This is the one case where that PR is **worse than the code it replaces** (both are far better than the raw-sum version in production today).

AR(1) flat ground, 30 km, marginal sigma 4 m, true ascent 0:

| spacing | correlation time | #382 | `main` before #382 |
|---|---|---|---|
| 1.5 m | 10 samples | **807 m** | 468 m |
| 1.5 m | 30 samples | **781 m** | 670 m |
| 5 m | 10 samples | **730 m** | 488 m |
| 40 m | 1 sample (white) | 59 m | 818 m |

A realistic 10 km flat hike recorded at 1 Hz walking pace (~1.5 m spacing): sigma 2.5 / tau 20 s gives 155 m against `main`'s 136 m; sigma 4 / tau 20 s gives 284 m against 218 m.

**Barometric-class noise is unaffected** (sigma 0.8-1, tau 60-300 s: 7-30 m either way), so this is specifically the phone-or-watch-without-a-barometer case — which is a common way to record a walk.

## Why the current design cannot see it

`main`'s band was `3 x raw sigma`, and `sigma` was under-measured the same way — but 3x an under-measured sigma still left a wider band than 4.5x a residual that assumes the noise averages away. It does not average away when it is correlated.

## Direction

The band needs to reflect how far the *smoothed* series actually wanders, not how far a white series of the same innovation would. Candidates:

- Measure the residual empirically on the smoothed series rather than predicting it from a formula — noting that the wander survives smoothing, so `raw - smoothed` is not it either.
- Estimate the marginal deviation from large-lag differences (lag >> correlation time captures total variance, ~`2 * sigma_marginal^2`) and floor the band on that — with the difficulty that on real terrain a large-lag difference also contains genuine elevation change, so it needs detrending.
- Detect correlation directly by comparing sigma estimated at several lags, and widen the window (rather than the band) when the ratio says the noise is correlated.

Worth building a small reproducible tuning harness in the repo as part of this — the constants in `track_edit.py` currently cite fixtures that live only in review transcripts, which is how the last round's figures ended up unreproducible.

## Acceptance criteria

1. AR(1) flat ground, 30 km at 1.5 m spacing, marginal sigma 4, tau 10-30 samples reports **< 100 m** (currently 780-810).
2. No regression on the cases #382 fixed: sparse-noisy flat 0 m, clean sinusoidal rollers ~500 of 600, held-altitude flat ~0, 1 Hz noisy climb ~599, sparse clean route 600.
3. Real ascent on correlated-noise input is still recovered — a 600 m climb recorded by a phone at 1 Hz must not be flattened by whatever widening this adds.
4. The tuning fixtures are committed and runnable, not quoted from a transcript.

Follow-up to #382 (issues #376, #374). Related: #260.

rui-nar:
## Summary

Closes #386 — the regression the #382 review found, where this code was *worse* than what it replaced for one common recording type.

The hysteresis band was sized from a white-noise model. k-th differences see only the **innovation** of an autocorrelated series, and GPS altitude error is strongly autocorrelated — satellite geometry and multipath drift over minutes, not samples. So an AR(1) series of marginal sigma 4 measured 0.5-0.9, the smoothing window stayed at its floor, the residual model concluded the noise had averaged away, and the band collapsed to its own floor. 30 km of flat ground came back as 700-840 m of climb.

That is a phone or a watch without a barometer, which is how most walks are recorded.

## The fix

Escalate the differencing stride. As the stride passes the correlation length the samples become effectively independent and the estimate saturates at the true marginal sigma — so keep doubling while the estimate climbs, and use the stride it saturated at. That stride is also the correct divisor for the independent-reading count the band derives from, so one change repairs both halves of the error.

| AR(1) flat ground, marginal sigma 4, true 0 | before | after |
|---|---|---|
| tau 10 @ 2 m spacing | 839 m | **0 m** |
| tau 30 @ 2 m spacing | 720 m | **0 m** |
| tau 10 @ 5 m spacing | 757 m | **0 m** |
| tau 30 @ 5 m spacing | 506 m | **44 m** |
| sigma 2.5, tau 20 @ 2 m | 448 m | **0 m** |
| real 600 m climb under AR(1) sigma 3 | 695 m | **600 m** |

## Why it has three brakes, and why it needs all of them

Terrain does not saturate — it grows without limit once the stride spans a hill. My first cut had only a span cap, and **it erased clean 200-300 m rollers outright: 483 m of real climb reading as 0.** That is precisely the failure #376 was filed for, reintroduced by the fix for #386, so it now has its own test.

- **Noise floor, 0.4 m** — below this there is nothing measurable to saturate, and escalating could only find terrain. A DEM-sampled route reads 0.23.
- **Growth ratio, 8x** — real saturation is bounded (`sigma_marginal / innovation` is `sqrt(tau/2)`: ~2x at tau 10, ~4x at tau 30). An estimate still climbing past that is terrain.
- **Span, 1 km** — past this a k-th difference spans whole hills.

Every terrain fixture is byte-identical after the change: clean sinusoidal rollers from 200 m to 600 m wavelength, noisy triangular rollers, the 1 Hz climb, the sparse planned route. 200k samples still cost 0.22 s.

## Honest remainder

`tau 30 @ 5 m spacing` still leaves 44 m over 30 km. The growth-ratio brake stops saturation short there, and loosening it far enough to close the gap is what started erasing rollers. Conservative and bounded beats inflated.

## Test plan

- [x] `pytest tests/` — **2764 passed, 2 skipped**
- [x] `test_correlated_noise_on_flat_ground_is_not_climbing` and `test_real_climb_under_correlated_noise_survives` confirmed **failing against `main`** (stashed the source, kept the tests)
- [x] `test_clean_rollers_survive_the_stride_escalation` passes on `main` by design — it guards this change's own failure mode, not the old bug
- [x] 200k-sample timing re-measured at two spacings

## Release notes

- [x] `Release-Note:` trailer under the `stats` scope


rui-nar:
## Adversarial review applied (`87feb9a`)

Verdict was **not safe to merge**, and it was right. The AR(1) numbers all reproduced, but the escalation had a second failure mode — and it is the one #376 was filed for, reintroduced by the fix for #386. Reproduced every finding before acting.

### The blocker: escalation latched onto under-sampled terrain

Past ~25 m between samples the terrain itself is under-resolved — 200 m rollers are four to eight points per hill — so its third differences are large for reasons that have nothing to do with a sensor. The escalation read that as correlated noise, inflated sigma to its ratio cap, and the band swallowed the hills whole:

| case | real climb | before the fix |
|---|---|---|
| clean 200 m/6 m rollers @ 50 m — **no noise at all** | 360 m | **0** |
| 200 m/10 m @ 25 m, white 0.5 | 405 m | **0** |
| 200 m/20 m @ 25 m, white 1.0 | 810 m | **0** |
| triangular 200 m/10 m @ 25 m | 334 m | **0** |
| 300 m/30 m @ 40 m, smart recording | 994 m | **0** |

Every failing case is at 25 m spacing or coarser, and not one fixture sampled that coarsely with enough noise to clear the floor — which is exactly why "every terrain fixture byte-identical" was both true and worthless as evidence. That is the more useful half of this review: the claim was not wrong, the fixture set was.

**Fix:** the escalation runs only where it is meaningful — a median gap of 15 m or less, which is what a fix-per-second recording gives (~1.5 m walking, ~7 m cycling, ~14 m at 50 km/h) and is the only thing that produces the correlated drift this is for. Coarser series are smart-recorded or simplified and carry few enough samples that accumulated drift is bounded anyway. Threshold chosen against the sweep, not by feel: a 5 m gate would have missed a 7 m-per-sample recording that AR(1) still inflates.

All five cases return to their previous figures; every AR(1) case stays fixed. New test covers all five, confirmed failing without the gate.

### F3 — mine, not #386's

Hoisting the median gap out of `_smooth_elevations` made a **zero** gap disable smoothing outright, where before it fell back to the span floor. More than half the samples sharing a position is a stationary stretch, and a 300 m climb recorded from a phone left sitting still came back as 534 m. Restored; now 299.

### F4 — the cost claim was best-case

"0.22 s per 200k" holds for white noise, which settles at the first stride and never enters the escalation loop. On the input this is actually for, AR(1) at 2 m spacing, it is **0.45 s** against `main`'s 0.16 s. Still linear. `test_long_track_stays_linear` measured only white noise, so it never exercised the escalated path — it measures both now.

### F2 — accepted and filed, not fixed here

The 0.4 m floor means the fix covers marginal sigma 4 up to tau ~55 samples; at tau >= 60 the result is byte-identical to `main` (500-650 m of phantom climb). That is the "drifts over minutes" regime the commit message leans on, so the Release-Note overpromises for slow-drifting sensors. It is **not a regression** — those inputs are exactly as they were — but it is unfinished, and the review showed removing the floor alone does not fix it either, because the ratio and span brakes then cap saturation below marginal. That needs a different mechanism, so it is a follow-up rather than another patch on this branch.

### F5 — test gap

Closed by the new coarse-spacing test; that is the one the review said would have caught F1, and it would have.

### Verification

- `pytest tests/` in the worktree — **2756 passed, 3 skipped** (a clean checkout skips `test_deploy_script.py`, which needs a gitignored local file)
- New coarse-spacing test confirmed failing without the gate
- Stationary-stretch regression re-measured: 534 -> 299 against a true 300

rui-nar:
## Stopping this PR — the cure costs more than the disease

Second review of `87feb9a` found a blocker I have reproduced independently, and it is not fixable by another constant. **Recommending this PR is not merged.** Converting to draft rather than merging.

### Rolling terrain under correlated noise is erased — inside the gate

200 m / 10 m rollers, 20 km, AR(1) noise. "clean" is what each version recovers from the same terrain with no noise at all:

| spacing | true | clean | main | this PR |
|---|---|---|---|---|
| 2 m, sigma 3 tau 20 | 1000 | 765 | 882 | **303** |
| 2 m, sigma 4 tau 10 | 1000 | 765 | 965 | **0** |
| 7 m, sigma 3 tau 20 | 990 | 748 | 793 | **48** |
| 14 m, sigma 4 tau 10 | 990 | 715 | 769 | **55** |

And it arrives as a cliff, not a slope — same road, same tau, marginal sigma varying by a quarter of a metre:

```
sigma 2.25   main 763   PR 763
sigma 2.50   main 773   PR  23
```

Two people riding the same road with slightly different phones get 763 m and 23 m.

### Why no constant fixes it

The escalation is working correctly — it measures sigma ~3 and a correlation length of ~100 m, and those numbers are right. The problem is what follows from them: suppressing correlated noise of sigma 3 with a 100 m correlation length requires averaging over roughly 800 m of travel, and that annihilates a 200 m roller. Capping the span at 240 m does not rescue it — 240 m still removes ~84% of a 200 m feature, and the band that the same sigma implies (9-12 m) then eats what is left of a 10 m hill.

**10 m rollers and 3 m correlated noise with a comparable correlation length are not separable from a distance-elevation series.** Their spectra overlap. Every gate I have tried moves which inputs get which error; none removes the trade.

Weighing the two errors: `main` over-reports flat ground (700-840 m over 30 km, bad) but stays close to its own clean baseline on rollers. This PR zeroes flat ground (good) and destroys up to 100% of real climb on ordinary rolling terrain (much worse — a user notices a 1000 m ride reading 0 far more than a flat walk reading 700).

### The one axis that would work

Noise correlates in **time**; terrain correlates in **distance**. That is the only place the two genuinely differ, and the elevation profile has no time axis. Per-point timestamps become available when #260 unit 3 lands its GPX time parsing — that is the honest path, and I have moved it into #386 rather than keep tuning constants against synthetic fixtures.

### Also found, and mine

**The PR head carries no tests for its own behaviour.** All four tests I added across `a250172` and `87feb9a` are absent from the tree. Cause: an edit that replaced the slice between `test_long_track_stays_linear` and `test_steady_climb_is_counted_in_full` — the four new tests sat inside that span and went with it. The full suite went 2764 -> 2756 at exactly that point, I noticed the drop, and attributed it to gitignored local files instead of checking. The reviewer was right to lead with it; a green suite proves nothing about tests that no longer exist.

That is a process failure independent of the algorithm, and the reason the second review was worth running even though CI was green.

rui-nar:
## Closing: measured, rejected, and superseded

This is not abandoned work — it was measured against the benchmark and turned down on the numbers, and those numbers are worth keeping visible.

**What it got right.** Escalating the differencing stride until the noise estimate saturates recovers the true marginal sigma of correlated GPS altitude error, which the stride-1 estimator underestimates because it measures the innovation instead. That is a correct measurement and the diagnosis behind it stands.

**Why it cannot ship.** What follows from the measurement is the problem. Suppressing noise of that size needs averaging over ~800 m of travel, which annihilates a 200 m roller, and the band the same sigma implies (9–12 m) eats what is left of a 10 m hill. On 20 km of 200 m/10 m rollers it reported **0–303 m against a true 1000**, where what shipped reports 769–965. It also arrives as a discontinuity rather than a gradient: the same road at marginal sigma 2.25 read 763 m and at 2.50 read 23 m, so two people riding together with different phones would get different trips. Full tables in [the withdrawal comment on #386](https://github.com/rui-nar/ViewTripWeb/issues/386#issuecomment-5641827516).

**Why not revive it later, either.** Under the direction #386 is now taking, this would make things *worse*: a correct-but-larger sigma widens both the smoothing span and the hysteresis band, and #412 is about those two already erasing relief the terrain model can see. Of a terrain model's 194–199 m on gentle rollers, the recording's 240 m span alone leaves 19–26 and its 5.5 m band alone leaves 0. A larger sigma deepens exactly that.

**Where the work lives now.**

- The input classes this PR was caught by are gates in `tests/elevation_bench/` — including the tau sweep that exposed it, which was a no-op at 60 s and a catastrophe at 10 s, so a benchmark fixed at either value would have missed it.
- #386 is being pursued by terrain substitution in #410: bring in an independent measurement rather than try harder to separate two signals that share a spectrum.
- The branch `fix/386-correlated-noise` (`87feb9a`) stays in the repository if the stride-escalation estimator is ever wanted on its own.

The lesson this PR paid for is on the issue: **10 m rollers and 3 m correlated noise of comparable correlation length are not separable from a distance–elevation series.** No constant fixes that, and three were tried before this one.

PR #393, opened 2026-09-11
fix(stats): benchmark the elevation classes, and keep the source's climb figure through an edit (#386 U0+U1)

rui-nar:
## Summary

**U0 of the elevation plan** — the unit every previous round lacked. Tests only; no behaviour change.

Elevation gain has been rewritten four times (#260, #371, #382, #389). Each round measured its change against whichever fixtures were to hand, shipped, and was then found to have wrecked a class of track nobody had thought to generate. A sample-counted smoothing window erased planned routes (600 m read as 3 m). A distance-counted one erased rolling terrain. A noise estimate that finally measured correlated drift *correctly* erased rolling terrain outright (1000 m read as 0) and had to be withdrawn.

Every one of those was cheap to detect and expensive to discover. This enumerates the input classes once.

## What it is

`tests/elevation_bench/` — generators plus a case table; `tests/test_elevation_bench.py` — the assertions. `python -m tests.elevation_bench` prints the table.

Cases are named by **recording class**, not by terrain shape, because what decides the figure's accuracy is which sensor produced the series:

| class | what it is |
|---|---|
| `clean` | a planner's terrain model — Komoot, RideWithGPS, a Garmin course. Sparse, exact, no sensor at all |
| `barometric` | a watch with a pressure sensor: small noise, slow weather drift. The good case |
| `phone` | GPS altitude, no barometer. Large error, correlated over minutes — the hard case |
| `white` / sub-sampled | a device recording one point every so often, or a track simplified before upload |

**21 gates** are asserted. Their bounds come from what the current implementation measures, so they pin behaviour against regression rather than claim accuracy — where a bound is loose it holds a *known* loss and the note says which (200 m rollers sit near the 60 m smoothing floor and lose ~25%).

**10 informational cases** are printed and never asserted. These are the ones this approach cannot get right, where the sensor's drift and gentle terrain are the same signal. Keeping them measured is more honest than leaving the gap inside a closed issue, and they become gates when terrain substitution lands. Current phantom climb on 30 km of flat ground, phone GPS: **471-822 m** depending on correlation time.

Two properties beyond the table: the figure must move **smoothly** as the sensor gets worse (#389 reported 763 m and 23 m for the same road a quarter-metre of noise apart), and splitting a track must not **manufacture ascent**.

## The part that makes it work

The phone cases sweep the noise's **correlation time** rather than fixing it.

That number is a convention, not a measurement from real handsets, and it decides how much of the error survives smoothing. My first draft of this benchmark fixed it at 60 s — and passed cleanly against the withdrawn #389, the exact regression it exists to catch. #389 is a no-op at 60 s and a catastrophe at 10 s. Sweeping is what turns these into gates that bite.

Rolling terrain under phone noise, true 1000 m:

| correlation time | main | withdrawn #389 |
|---|---|---|
| 10 s | 974 | **0** |
| 20 s | 953 | **0** |
| 30 s | 912 | **163** |
| 60 s | 853 | 853 |
| 120 s | 818 | 818 |

## Validation — it catches every regression in the history

Run against each version of `track_edit.py` in turn:

| version | result |
|---|---|
| raw positive-delta sum (pre-#371) | **9 of 21 gates fail** — flat barometric 3449 m, flat white 33724 m, a 600 m climb 2367 m |
| `96b6dfc` sample-counted window | **8 gates fail**, including the planned routes |
| `ae990d4` its successor | **4 fail**, including the no-cliff property |
| `fix/386-correlated-noise` (withdrawn #389) | **3 phone-roller gates fail** |
| `main` | **all 31 pass** |

## Test plan

- [x] `pytest tests/` — **2798 passed, 3 skipped**
- [x] Benchmark validated against four historical versions as above
- [x] `python -m tests.elevation_bench` prints both tables

## Release notes

- [x] No user-visible change — tests only, subject line is enough

Part of #386. Next: U1 (store the source figure and apportion edits/splits/resets).


rui-nar:
## U1 pushed (`2513fc1`) — keep the source's figure through an edit

### The defect, restated in what a user sees

Open the track editor on a Strava ride, save without touching anything, and the climb changes. Trim fifty metres off it and it can move by hundreds. Split it in two and the pieces do not add up: on this repo's own split fixture, a 60 m activity became two pieces totalling **39 m** — 21 m of climb deleted from the trip by cutting it in half.

The cause is that an edit re-measured the gain from scratch, discarding Strava's own figure — measured from data we never receive and corrected in ways we cannot reproduce — and substituting ours.

### The change

An edit now **scales** the existing figure by the share the new geometry accounts for, instead of re-measuring it. That is the treatment moving and elapsed time have always had, a few lines below in the same function, for the same reason.

The share uses our own estimator on *both* sides, so its absolute bias divides out and only the proportion survives. Two things fall out of that:

- **The split needs no special case.** Both pieces are already apportioned against the same pre-split denominator, so they sum to the parent.
- **The estimator's accuracy stops mattering here.** It only has to be consistent, which it is.

One new column, `original_total_elevation_gain` — the snapshot of what the number was before the first edit — living and dying with the `original_polyline` and `original_elevation_profile_json` beside it. Written on the first edit, restored and cleared on reset.

**No backfill.** NULL means "never edited, nothing to restore", and the edit path falls back to recomputing when it finds no snapshot, which is what every row did before today. That same fallback covers E2EE activities, whose stored profile is ciphertext to the server so there is no 'before' geometry to take a share of — tested explicitly.

### Deviation from the plan, and why

The plan called for a `source_total_elevation_gain` column backfilled by guessing which rows were pristine (`source IS NULL AND is_edited = 0`), plus separate apportioning code in the split and reset paths.

Working through it, that turned out to be more machinery than the problem needs. Scaling the *current* stored value composes correctly on its own: on a first edit the stored value **is** the source figure, and on later edits it is already the correctly-scaled one, so nothing needs to know which is which. That removes the backfill heuristic entirely — and with it the risk of mislabelling a row — and leaves the split falling out of the shared denominator rather than needing its own arithmetic. The one thing genuinely worth storing is the reset snapshot, which is exactly the shape of the columns already there.

### What this does *not* do

Nothing about what the estimator measures. #386's phantom climb on phone recordings is untouched — that is U3 (terrain substitution). This stops the app discarding a better figure than its own wherever one already exists.

### Verification

- [x] `pytest tests/` — **2808 passed, 3 skipped**
- [x] Both new integration tests confirmed **failing without the change**: the split pieces summed to 39 of 60, and no snapshot was taken
- [x] 8 unit tests on the rule itself, including the no-op edit, the compounding second edit, and the unreadable-profile fallback
- [x] `alembic upgrade head` from an empty DB, then `alembic check` — no drift
- [x] The U0 benchmark still passes unchanged, which is the point of having it

PR #395, opened 2026-09-12
feat(gpx): recognise a re-imported file, and allocate local ids one way (#260 unit 2)

rui-nar:
## Summary

**Unit 2 of the #260 plan** — give an imported activity an identity, and stop two code paths inventing their own.

### Re-importing a file made a second activity

An activity's `id` says which row it is. For a Strava sync that is also *which real-world activity* it is — the id **is** Strava's id, so the upsert is idempotent by construction. A GPX import had no such handle: it took a local id drawn at random, so importing the same file twice produced two overlapping activities, doubled the trip's distance and climb, and left nothing for the user to notice it by.

`source_id` is the missing handle: a fingerprint of the imported track, hashed from its coordinates and start time.

Three decisions worth stating:

- **Scoped to the trip, not the table.** The same commute ridden on a tour, or a loop walked on two holidays, legitimately belongs to more than one trip. "Already imported" is a question about one timeline. The column is deliberately *not* uniquely indexed.
- **The start time is hashed too.** Geometry alone would call two laps of the same loop one activity.
- **Elevation is not.** It is the part of a track most likely to be rewritten after import — by an edit, or by the dropout repair in `c4a9e1f70b38` — and a fingerprint that changed when the app corrected its own data would stop recognising the file it came from.

A refusal names the activity already holding the track (`409` with its `activity_id`), so unit 5's dialog can offer to open it rather than only saying no.

### Two allocators became one

| path | scheme |
|---|---|
| split tail | `min(0, global_min) - 1` — dense, deterministic |
| GPX import | `-secrets.randbits(62)` — sparse, random |

Two schemes sharing one keyspace with different collision properties. The decrementing one is unsafe once ids travel between deployments: it hands out -1, -2, -3 in *every* one of them, and a `.viewtrip` exported from one carries its ids into another, where `activity` is keyed by id alone and ingest upserts.

Both now use `allocate_local_activity_id`, which keeps the property that mattered in the split path — it checks the activity **table**, not a project's timeline, because a tail whose item was removed leaves the row behind and reusing its id collides on INSERT.

### Two existing tests were asserting the old scheme

`test_split_across_projects_allocates_distinct_ids` asserted `tail1 == -1` and `tail2 == -2` — the particular values the old allocator produced, not the property it was written for. `test_synthetic_ids_do_not_collide` imported the same bytes twice to prove ids differ, which is now a duplicate rather than a second activity.

Both now assert the intent: negative and distinct, with distinct geometry for distinct activities. They keep testing what they were about.

### No backfill

Existing GPX imports keep a NULL fingerprint. The bytes they came from are gone, and one derived from the stored geometry would not match what the importer computes from a file — that geometry has been through interpolation and repair since. They do not participate in duplicate detection, which is exactly what they did before this column existed.

## Test plan

- [x] `pytest tests/` — **2822 passed, 3 skipped**
- [x] New: 12 unit tests on the allocator and the fingerprint (exhaustion, collision avoidance, 64-bit bounds, float round-trip stability, order sensitivity, same-route-different-day)
- [x] New: re-import refused and names the existing activity; the same track accepted into a *different* trip
- [x] `alembic upgrade head` from an empty DB, then `alembic check` — no drift

## Release notes

Part of #260. Next: unit 3 (parse layer — `<rte>` support, timestamps, derived name and type).


rui-nar:
## Adversarial review applied (`d0f01bb`)

Verdict was **not safe to merge**, on one finding, and it was right — and it turns out to describe a bug that is already live on `main`, not just one this PR would have introduced.

### Local ids were wider than the web client can carry

The web build is dart2js, where Dart's `int` **is** an IEEE-754 double. An id beyond 2^53 is silently rounded the moment `jsonDecode` sees it, and the client sends that rounded value back on every edit, split and delete — so the server 404s an activity the user is looking at.

Verified here rather than taken on trust:

```
flutter build web --release          (docker-build.yml:47 — dart2js, no --wasm)
activity_editor_page.dart:62         (widget.activity['id'] as num).toInt()
62-bit ids: 5/5 change when passed through a double
            -3021798623442944499 -> -3021798623442944512
53-bit ids: 0/5 change
```

**`main` already ships this.** GPX import has drawn 62-bit ids since it shipped, so deleting or editing a GPX-imported activity has been broken in the browser all along — silently, because Android and iOS carry real 64-bit ints and never showed it. Unifying the allocators is what lets one constant fix both; without the unification this PR performs, splitting would have inherited it instead.

`_LOCAL_ID_BITS = 53` keeps every id inside `Number.MAX_SAFE_INTEGER`, and the collision odds stay negligible — about n²/2^54, one in eighteen million at a million local activities — against the decrementing scheme it replaces, which handed out -1, -2, -3 in *every* deployment and so collided with certainty once a `.viewtrip` crossed between two.

**The test that claimed to bound this asserted the wrong limit.** `test_stays_inside_a_signed_64_bit_column` checked `> -(2**63)` — the database column's bound, not the one that actually binds. It now asserts the JS-safe range and that each id survives a round trip through a double.

### Documented rather than changed

**The mistyped-time gap (F2).** The fingerprint's start time is whatever the importer is given, which today the user types — so the same file imported twice with a typo'd time is not recognised and lands as a second activity. Reproduced by the reviewer. The alternative, ignoring time, merges a planned route genuinely ridden on two days, which is a real case. Unit 3 takes the time from the file where it has one, narrowing this to files carrying none. Now stated in `track_fingerprint`'s docstring rather than left implicit.

**`.viewtrip` can carry a duplicate fingerprint past the check (F3).** That path upserts by id and never consults `source_id`; dedupe guards the GPX endpoint only. Not a regression, and out of this unit's scope.

**The 409 message can show an E2EE ciphertext name (F4).** Cosmetic — unit 5 will use the `activity_id` field, not the prose.

### What the review confirmed unbroken

Worth recording, since these were the parts I was least sure of: the dedupe query handles an activity referenced twice in one project, an orphaned row, and a different project correctly; the check-then-insert race is unreachable (synchronous DB block, single uvicorn worker); the fingerprint is stable across XML serialisation variants and float round trips; nothing orders on activity id magnitude, so randomising split-tail ids breaks no split-family ordering, renumbering or pruning; and `test_split_across_projects_allocates_distinct_ids` **still catches the original cross-project id-reuse bug** after being loosened — which was the specific thing I asked about, having weakened the assertion myself.

### Verification

- [x] `pytest tests/` — **2822 passed, 3 skipped**
- [x] dart2js precision claim reproduced directly (see above)
- [x] Reviewer ran the new dedupe test against an `origin/main` extract: fails `200 != 409`, so it detects the regression

PR #396, opened 2026-09-12
feat(gpx): import planned routes, and read what the file already says (#260 unit 3)

rui-nar:
## Summary

**Unit 3 of the #260 plan** — the parse layer. Pure: no HTTP, no database, no clock, so unit 4 can build the inspect/preview step on top of it.

> Stacked on #395 (unit 2). Review that first; this diff is `7179a1d` alone.

### A planned route is importable

Issue #260 was written for *"a route drawn in Komoot or RideWithGPS"* — and those planners frequently export a `<rte>`, not a `<trk>`. Reading only `gpx.tracks` meant the motivating case came back as **"GPX contains no tracks"**, which is both a refusal and a lie about why.

- Routes are read when a file has no track.
- A file carrying **both** still imports the recorded track — Garmin courses often include the route they were planned from, and what happened beats what was intended.
- A waypoints-only file now says *that*, instead of claiming it has no tracks.

### The file's own facts are read instead of retyped

A GPX carries the date, the start and end time, usually a name and often a type. The first version ignored all of it while asking the user to type it in for a ride they did three weeks ago.

| fact | where it now comes from |
|---|---|
| date, start, end | `<time>` on the points |
| moving time | elapsed minus the standing still |
| name | track `<name>` → file metadata `<name>` → filename stem |
| activity type | `<type>`, mapped through a table of what real tools write |

**Moving time was previously set *equal* to elapsed time**, so every import claimed it had never stopped even when its own timestamps disagreed. A sample under 0.3 m/s is not moving; and a gap over five minutes is the device switched off rather than a slow stretch — without that second rule a two-hour lunch lands in the ride's moving time.

**The name matters more than it looks.** The filename stem is the weakest of the three sources and is what the first version *always* used, which is why imports were called things like `2024-08-12_073312`.

**Unrecognised types stay unset** — including the bare numbers some Garmin exports write — so the user is asked rather than told a kite surf is a hike.

### A multi-track file is described rather than dismissed

Each candidate reports its name, point count, distance and timing, so unit 5's picker can offer them; `validate_for_import` takes a `track_index` to check one. Without an index the file is still refused — choosing would be guessing which ride was meant — but the message says a choice exists instead of *"only a single track is supported"*, which left the user to go and edit the file in another tool.

### Uploads are size-guarded before parsing

gpxpy builds an object tree many times the size of the XML — a 4.6 MB file measured at 73 MB of heap — so a point cap applied *after* parsing is a cap that has already let the damage happen. The guard is exported here and wired up in unit 4.

### Three existing tests asserted the old wording

`"no tracks"`, `"only a single track"`. The behaviour they cover is unchanged — those files are still refused — so they now assert the new messages. Flagging it because a wording-only test change is exactly where a behaviour change can hide, and this one isn't: the multi-track and zero-track paths still return errors, and the API test still expects 422.

## Test plan

- [x] `pytest tests/` — **2854 passed, 3 skipped**
- [x] 32 new tests: route-only files, route-vs-track precedence, waypoints-only, times and moving time (including a rest and a switched-off device), name precedence across all four sources, type mapping and its refusals, multi-candidate description and per-index validation, the size guard
- [x] Existing GPX suites green with only the three wording assertions updated

## Release notes

Part of #260. Next: unit 4 — the inspect endpoint and import v2, which is where these derivations reach the user (and where elevation U2's "estimated" label lands).


rui-nar:
## Adversarial review applied (`50e61ef`)

Verdict was "mergeable, with two things to fix before unit 4 builds on it". The first one made this PR's release note **false**, so it was worth more than its severity rating suggested.

### Moving time counted stops as movement

"The moving time no longer counts the stops" held only for a device with **auto-pause**, which emits no samples while you stand still. A phone recording continuously emits a position that wanders — and a per-sample speed test is a displacement test in disguise: at 1 Hz, 0.3 m/s is 0.3 m *between neighbours*, which ordinary jitter clears without the device going anywhere.

Reproduced on a twenty-minute stationary recording, share counted as moving:

| jitter | per-sample rule | 30 s window |
|---|---|---|
| white 0.5 m | **91%** | 0% |
| white 1.5 m | **99%** | 0% |
| drift 2.5 m | **93%** | 5% |
| *a real walk at 1.4 m/s* | 90% | **100%** |

Speed is now net displacement across 30 s of track. A random walk goes nowhere, so jitter cannot fake it.

**One subtlety the existing fixture caught while I was writing this.** Judging by a window but crediting the sample interval lets a fast window authorise a long stationary gap: a two-minute rest between two walked samples got counted whole, because its neighbours were moving. An interval longer than the window carries its own verdict instead — over that much time, real movement dwarfs jitter. That test was already there from the first commit and failed immediately, which is the only reason I noticed.

### An empty `<trk>` suppressed a real `<rte>`

This is the edge I flagged when launching the review, and it was real. Some tools write a placeholder track beside the route; taking it as a track hid the route, and the file was then refused for *having a track with no points* — true, and useless. Point-less elements are no longer candidates, and what remains is renumbered so an index still addresses the list a picker was offered.

### Garmin Connect's type vocabulary was missing

I built the alias table from plausibility, and it showed: Garmin writes `road_biking` and `trail_running`, while the table had only the space-separated spellings **no verified tool emits**. Separators are normalised now, so one table covers Strava's `cycling` and Garmin's `road_biking` alike.

### Three more

- `elapsed_seconds` took first and last stamps rather than earliest and latest, so a device's occasional backwards clock step reported a span shorter than the ride — or none at all. Moving time is clamped to it too, since a preview showing more moving than elapsed is visibly wrong.
- A file mixing zoned and unzoned `<time>` values raised out of the middle of a duration calculation — downstream, a 500 rather than a refusal the user can act on. Naive stamps are read as UTC, which is what the spec says they are.
- `gpx_track_to_points(-1)` quietly returned the last candidate; it refuses now.

### Not changed

**`guard_upload_size` is still dead code** — wired up in unit 4, as the PR said. The reviewer measured the docstring's claim and it holds (4.8 MB → 73 MB peak, 15.4x), but also found a Garmin-style file with `<extensions>` costs 261 B/point, so 20 MB is ~80k points rather than ~200k. Worth revisiting the constant in unit 4 when it actually guards something.

### Verification

- [x] `pytest tests/` — **2871 passed, 3 skipped**
- [x] 17 new tests covering every finding; **all confirmed failing against the previous commit** (`git stash` on the source alone, tests kept)
- [x] The stationary/walking figures above re-measured after the fix

PR #402, opened 2026-09-12
feat(gpx): read a GPX file before importing it, and fill the form from it (#260 unit 4)

rui-nar:
## Summary

**Unit 4 of the #260 plan** — the dry run, and an import that trusts the file.

> Stacked on #396 (unit 3), which is stacked on #395 (unit 2). Review those first; this diff is `3429586` alone.

### `POST /{name}/activities/gpx/inspect` — reads an upload, writes nothing

That is the whole feature. Writing nothing is what lets the client show a preview, prefill every field from the file, offer a choice between several tracks, and warn that a track is already in the trip — all *before* anything is created. The first version asked for a date and a time up front and only then said whether the file was acceptable at all.

Per candidate it returns: name, mapped activity type, point count, distance, whether it is a planned route, its timestamps, elapsed and moving time, the estimated climb, and **its own rejection reasons**, so the client can grey out what it cannot import rather than discovering it on submit.

### The import's form fields are now optional

What is omitted comes from the file — date and times from `<time>`, name from `<name>`, type from `<type>`. They remain *accepted* because the file does not always have them (a planned route carries no clock) and because the user is entitled to correct what it does say. A form value always wins.

### Elevation plan U2 lands here

The climb is reported as `elevation_gain_m` with `elevation_gain_estimated: true`, because the app measures that figure itself rather than being told it, and every surface that shows it should say so.

It needed **no new storage**: unit 1 of the elevation plan made the stored figure the basis an edit apportions from, so computing it once at import already *is* "store it as the source figure". The plan asked for a `source_total_elevation_gain` column here; that turned out to be unnecessary for the same reason it was unnecessary in U1.

### Two bugs found while writing the tests, both mine

- **`track_index` defaulted to `0`**, so a multi-track file silently imported the first track — the *silent partial import* #260 explicitly ruled out. It defaults to unset now; omitting it on a multi-track file is still a refusal.
- **The moving-time guard replaced any value outside `0 < t <= elapsed` with the elapsed time**, which turned a legitimately **zero** moving time back into "never stopped" — undoing what unit 3 had just fixed. It clamps rather than replaces, and there is a test asserting a stationary track reports zero.

### The size guard is now load-bearing, so the limit changed

20 MB → 12 MB. Unit 3's review measured the encodings: plain trackpoints cost ~95 bytes each, Garmin's `<extensions>` ~261, and the heap is ten to fifteen times the file. 20 MB let a rich 80k-point file through to be parsed at ~200 MB and **only then** rejected by the 50k point cap. 12 MB is ~46k rich points — under the cap, which is the point of having it.

### The project's own invariant test caught the new route

`test_every_mutating_project_route_busts_the_payload_cache` flagged `inspect` as a POST under `/api/projects` that does not bust the payload cache — correctly, by its heuristic. Exempted with a reason, because busting a cache here would be the giveaway that the dry run is not dry.

## Test plan

- [x] `pytest tests/` — **2888 passed, 3 skipped**
- [x] 17 new tests, **all 17 confirmed failing without the change**: inspect writes nothing (row count unchanged), a viewer is refused, the file's facts are reported, a route reports no clock, several tracks each get a verdict, a duplicate is flagged, an import needs nothing typed, a stationary track reports zero moving time, a form value overrides the file, a route without times is refused until given a date, a multi-track file needs a choice, a chosen track imports, an oversized file is refused before parsing

## Release notes

Part of #260. Next: unit 5 — the client flow, where this preview becomes a screen.


rui-nar:
## Adversarial review applied (`2adffa8`)

Verdict was **not safe to merge**, on a finding I would not have found: the dry run froze the server.

### `inspect` held the event loop for seconds per call

It is `async def` and did its whole parse and metrics pass synchronously. Reproduced here:

```
 8,000 points: 0.24s of synchronous work
20,000 points: 0.56s
40,000 points: 1.25s   (an ordinary ride)
```

During that window the instance serves **nobody**. Import had the same defect, but it fires once per import; `inspect` fires on every file pick, has no rate limit, and any editor on any trip could serialise the box at seconds per POST. And the size guard I made "load-bearing" one commit ago bounds **memory, not CPU** — so tightening it to 12 MB did nothing for this.

Both routes hand the synchronous half to a worker thread now, which is what the upload paths in `journal.py` and `memories.py` already do — these two were the only `async def` handlers in the module doing their own CPU work. There is a test that races a ticker against the call and fails if the loop stalls; it catches a **0.90 s** stall on the old code.

### One file had two fingerprints

The duplicate check hashed the *supplied* start time. The form carries only `HH:MM`, and devices start recording mid-minute — so importing a file from the preview (`07:33`) and then again with the form untouched (`07:33:12`) produced two identities, and **the second copy sailed straight past the check**. That is precisely the hole unit 2 added the fingerprint to close.

```
file time  07:33:12 -> b23dbe5aacfb624b
typed time 07:33:00 -> 89eb3d658eef09c1
```

The file's own start time is the basis now whenever it has one. The supplied date remains the basis for a file with no clock, where it *is* the distinguishing fact — the same planned route on Monday and Tuesday is still two activities, and there is a test for that. A fingerprint that cannot be judged is `None`, meaning "unknown" rather than "no duplicate".

### The preview promised what the import refused

`elapsed_seconds` took earliest and latest stamps while the import took first and last, so a device's backwards clock step gave a preview that prefilled happily and a submit that answered *"the timestamps do not cover a period of time"*. Both go through one accessor now — which matters more here than elsewhere, because this unit exists specifically to stop the user meeting a failure after filling a form.

### Four more

- A candidate that cannot be imported is no longer measured: a 129k-point file (which 12 MB still admits for plain trackpoints) cost 4.5 s to describe before its "too many points" verdict.
- The guard consults the upload's **declared** size before reading a byte. It ran after `file.read()`, so an oversized body grew the process by its whole length before being refused.
- A non-UTC `<time>` offset reached storage verbatim, against the column's documented ISO-8601 UTC — which also gave two exports of one ride (`05:33Z` and `07:33+02:00`) two fingerprints.
- A partial form on a file that *has* timestamps said the file has none. It now says the three fields go together.

### Two of my seven tests did not fail on the first attempt

Worth stating, since a test that passes before and after guards nothing. Both were my error: the loop test cancelled its ticker *before* it could wake to record the stall it had just sat through, and the backwards-clock fixture put its last stamp out of order but not actually before the first. Fixed, and all seven now fail against the previous commit.

### Accepted as-is

**`"Workout"` when the file's `<type>` is unrecognised.** The review is right that it sits oddly beside the parse layer's "return None so the user picks" principle. But the API needs *some* type, `activityTypeBucket` maps it to `other`, and it is a real Strava type — so it renders correctly. Unit 5 will always send a type; the fallback is for API callers that do not.

### What the review confirmed unbroken

`inspect` really writes nothing — checked against a dump of all 30 tables, the geo caches, and temp files, not just an activity count. The `_EXEMPT` entry is honest. The moving-time clamp holds end to end, including `average_speed` with a zero. Every form combination returns 200 or 422, never a 500. And all 17 original tests fail on the parent commit.

### Verification

- [x] `pytest tests/` — **2896 passed, 3 skipped**
- [x] 7 new tests, all confirmed failing against the previous commit
- [x] The stall reproduced and re-measured after the fix

Issue #365, opened 2026-09-10
GPX import: per-activity timezone, derived from the start coordinate (#260 F7)


rui-nar:
## Problem

An imported activity's timezone is asserted, not modelled. `import_gpx_activity` sets both `start_date` and `start_date_local` to `datetime.combine(day, clock, tzinfo=timezone.utc)` and writes the literal string `"UTC"` into `Activity.timezone` (`api/activities.py:396-425`).

For a trip in Japan, a user entering "09:00" gets a `start_date` of 09:00Z — nine hours off in every UTC-ordered comparison — while the local label happens to look right. Days mixing Strava-synced and GPX-imported activities can therefore sort wrong.

Split out of #260 (finding F7): v2 deliberately keeps today's behaviour rather than half-fixing it.

## The model that survives

A **per-activity** timezone derived from the track's start coordinate:

- It is what Strava already gives us for synced activities, so the two sources would finally agree.
- A trip crossing timezones is the normal case for this app, not the exception — so a single trip-level timezone field would be wrong, not merely coarse.
- GPX timestamps are UTC by spec, so coordinate + UTC instant is enough to resolve the local offset exactly, with no question put to the user.

## Cost to weigh before starting

- Needs a coordinate → zone lookup. Nothing in `requirements.txt` does this today (`zoneinfo` is stdlib but needs a zone *name*; `src/services/hafas_service.py` gets its names from HAFAS). `timezonefinder` carries ~40-50 MB of boundary data — check the image-size impact against the Docker build before committing to it.
- Needs its own backfill decision for existing GPX-imported rows.

## Acceptance criteria

1. A GPX imported into a trip abroad stores a `start_date` whose UTC instant is correct and a `start_date_local` that matches the wall clock at the track's start.
2. `Activity.timezone` carries a real zone name, not the literal `"UTC"`.
3. Existing GPX-imported rows are either corrected or explicitly left alone, with the choice recorded here.
4. Tests cover a trip spanning two timezones.

PR #404, opened 2026-09-12
feat(gpx): the file speaks first — pick, review, confirm (#260 unit 5)

rui-nar:
## Summary

**Unit 5 of the #260 plan** — the flow the four units before it were for.

> Stacked on #402 (unit 4) → #396 → #395. Review those first; this diff is `f4238fc` alone.

The old dialog asked for a date, a start time, an end time and an activity type up front, uploaded, and *only then* said whether the file was acceptable at all. A user filled in four fields for a ride they did three weeks ago and learned afterwards that the file had two tracks in it.

Picking a file now inspects it — the dry run from unit 4, which writes nothing — and the review step opens with everything the file already knew filled in: name, type, when it happened, distance, climb, and the track **drawn as a thumbnail** so "wrong file" is obvious while that still costs nothing. Every field stays editable, because the file is not always right.

### A test per state, because each state is a way the old dialog failed someone

| state | what it does now |
|---|---|
| reading | progress while a large file is parsed |
| several tracks | a choice, each with distance, climb and duration; unusable ones greyed **with the reason** rather than hidden |
| one usable among several | no question asked — straight to review |
| planned route | no clock, so date and times are empty and marked required rather than invented |
| duplicate | named *before* the import, not refused after it |
| date outside the trip | flagged, not blocked — importing the ride from the day before a trip is legitimate; silently extending the trip is not |
| rejection | back to the pick step with the server's own reasons |
| success | hands back the activity so the caller offers **View** and **Undo** |

### Three details worth their lines

**Untouched times are not sent back.** The form carries only `HH:MM` and the file's start has seconds in it, so echoing the truncated value would change the fingerprint the server dedupes on — and a re-import would not be recognised. This is the unit-4 contract trap the reviewer flagged, honoured on the client side.

**The type list gained "Other".** A kayak or a ski tour could previously only be imported as a mislabelled hike.

**The picker asks for `FileType.any`, not `custom` + `gpx`.** Custom greys out files whose UTI iOS does not recognise, and `.gpx` arriving from Mail or iCloud commonly is one — which may well mean GPX import has been unusable on iOS since it shipped. The server is the only thing that can really judge a GPX anyway.

Plus: the date picker aims at the trip rather than at today (a trip edited in February may be last summer's), and the dialog is sized to the viewport since a fixed 420 clipped it on a narrow phone.

### Server side

`inspect` gained a thinned preview outline per candidate, so the thumbnail has something to draw: stride-sampled to at most 200 points, first and last always kept. Stride rather than Douglas-Peucker because this is a thumbnail — a predictable point count in one pass, not the minimal set within a tolerance. A 50k-point track encodes to ~250 KB; this to about 1.

### Three of these tests failed on their first container run, all three my error

Worth stating plainly, since it is the second time this session:

- `MockClient` hands its handler a **materialised `Request`**, not the `MultipartRequest` that was sent, so `as MultipartRequest` threw. Fields are read out of the encoded body now.
- An outer `MediaQuery` is **discarded** — `MaterialApp` builds its own from the view — so the narrow-phone test was measuring an 800px window. It uses `setSurfaceSize`.
- The "unusable track" test asserted a choice the dialog **rightly does not offer**: with one importable track it skips the question. The premise was wrong, not the code; it now covers both shapes.

## Test plan

- [x] `flutter test` in the pinned container — **1480 passed** (16 of them this dialog's)
- [x] `flutter analyze` — no issues
- [x] `pytest tests/` — **2896 passed, 3 skipped**

## Release notes

Part of #260. Next: unit 6 — provenance surfaces (badge tooltip and semantics, the editor header, a source filter).


rui-nar:
## Adversarial review applied (`641691d`)

Verdict was **not merge-ready**, on three findings a user would hit without understanding why. All six actionable ones are fixed; the seventh is filed as #405.

### Touching any date or time field moved the activity by hours

The model converted the file's times to the device's local zone — which is the right thing to *display*, 09:33 for a 07:33Z ride under CEST — but the form posts `HH:MM` back and the server reads that as **UTC**. So opening the End picker and pressing **OK without changing anything** stored the ride two hours late, with nothing on screen looking wrong.

Times are kept in UTC now: displayed and submitted on one clock, until there is a real per-activity timezone to hold (#365).

**The test for this asserts `isUtc`, not the hour.** The container runs in UTC, where the hour alone looks correct — a test checking `startedAt.hour == 7` would have passed the whole time the bug shipped. That is the shape of test this defect needed.

### A ride over midnight could not be imported at all

Worse than it sounds, because **no user action was required**: the file's own span of 22:30 → 01:10 hit a minutes-past-midnight comparison and was refused while the form still said *"read from the file"*. And the user could not work around it — supplying times put both clocks on one date, which the server also rejected.

The server rolls an end before the start onto the next day now, the dialog says *"ends the next day"*, and only an identical start and end is wrong.

**That reverses one of my own server tests.** `test_end_time_before_start_time_rejected` asserted the old refusal — it was encoding the bug. It is now two tests: a night ride imports, and a zero-length one does not.

### A network error left a spinner turning with Cancel disabled

```dart
_stage = _fileBytes == null ? _Stage.pick : _stage;   // never null here
```

`_send` returns early when `_fileBytes` is null, so at that line it cannot be — the assignment was a no-op and the stage stayed `reading` or `submitting`, where Cancel is disabled. The only way out of the dialog was tapping the barrier. It returns to a stage the user can act from, and there are tests for both the inspect and the import path.

### Three more

- **A file whose only track is unusable** opened the review step on it: a form that could never be submitted, with the reason nowhere on screen. The reasons now appear where the file was picked — which is what the PR body already claimed happened.
- **An unrecognised `<type>`** left Import dead with no explanation, while every other missing field says "Required". The dropdown now says what it needs.
- **`_preview_polyline`'s stride used floor division**, so 399 points passed a cap documented as 200 completely untouched. Ceiling division, and the final point is appended only when the stride actually missed it.

### Filed rather than fixed here: #405

Undo can delete through the wrong trip, because `delete_local_activity` never checks the activity belongs to the project it was called for — it validates only that the id is local and the row exists. Pre-existing server laxity; this PR created the reachable path (the snackbar outlives the screen and the notifier's ref is read at tap time). **The client now checks the ref before deleting**, so the path is closed, but the server gap is real and #405 carries it — including the case that makes it more than a one-line fix: an orphaned local row, which `d5b1c0a2e3f4` documents as a genuine state, must stay deletable.

### Downgraded from the interim note

The painter's NaN assertion is unreachable — the server 500s on encoding a NaN polyline before it could be sent — and the model only throws on JSON *types* the server never emits. Neither is a finding.

### Verification

- [x] `flutter test` in the container — **1496 passed**
- [x] `pytest tests/` — **2897 passed, 3 skipped**
- [x] `flutter analyze` — no issues
- [x] 10 new tests; all 7 dialog ones confirmed failing against the previous commit

